Issue #4, opened 2019-03-28
HTTP status code can't be overwritten by route handler


nschlehe:
Generated controller class sets ctx.status after calling the registered route handler and overwrites previously set status code. An easy fix would be moving the status() call before the handler turning the behaviour to a fallback.

```java
ApiBuilder.post("/login", ctx -> {
      controller.login(ctx);
      ctx.status(201);
});
```

rbygrave:
Fixed with the provided PR - thanks !!

PR #5, opened 2019-03-29
#4 Set response status before calling route handler to avoid overriding status set in handler

nschlehe:


rbygrave:
Yes, makes good sense.  Apologies for being late to this. Thanks.

PR #26, opened 2020-05-05
Added asFloat and fixed a type LongHander --> LongHandler

Tijs-2:
I was using the library and had an issue with floats and I have tried to add them to the code

Please have a look,
Tijs


rbygrave:
Looks all good. Thanks.

PR #27, opened 2020-05-15
Added the @Tag and @Tags functionality.

Tijs-2:
Hi Rob,

I wanted to use tags so I tried to add the functionality and you can now add a one or multiple tags to a controller or to the method itself. They will be added instead of overwritten if you add multiple.

Small example how to add them:
@Tag(name="ControllerTag")
@Controller
public class Controller {

  @Tag(name = "EndpointTag")
  @Get
  public void get(Long id)  {}

  @Tags({
      @Tag(name="tag 1", description="Tag 1"),
      @Tag(name="tag 2", description="Tag 2")
  })
  @Post
  public void post() {}
}

I must admin not sure if this is the correct way to implement it but it seems to work :)

Tijs


rbygrave:
Yup, looks good - merging.

PR #50, opened 2020-12-30
Auto add exceptions

Tijs-2:
Hi Rob,

I have added something what could be a nice feature.
Auto adding a 404 status result for example in the openapi json file.

When you create an exception like this

    package com.example;
    
    import io.avaje.http.api.BadRequestResponse;
    import io.avaje.http.api.StatusCode;
    
    @StatusCode(505)
    public class ExampleErrorResponse extends BadRequestResponse {
      public int example= 1;
    
      public ExampleErrorResponse (String message) {
          super(message);
      }
    }

Then add it to a method like this:
     @Post
     public void updateExamplePost() throws ExampleErrorResponse

It will add it to the openapi with statuscode 505

Then to return these errors with the correct error code it should also be registered with the javalin service like this:

        javalin.exception(BadRequestResponse.class, (badRequestResponse, context) -> {
            context.json(badRequestResponse.getResponse());
            StatusCode statusCode = badRequestResponse.getClass().getAnnotation(StatusCode.class);
            if(statusCode != null)
                context.status(statusCode.value());
            else
                context.status(400);
        });

Hopefully you think it is a useful feature because I think it is quite handy to register fails also in the openapi documentation.

There are a few items left (or at least a few items)
- I do not know if the tests still run I cannot run them on my machine.
- I do not now if you can register the throwable exceptions the same way in helidon, jex etc because I have no experience with them.
- Not sure if Kotlin has the same way of throwing error so also not sure if this works

Tijs

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/BadRequestResponse.java
+++ b/http-api/src/main/java/io/avaje/http/api/BadRequestResponse.java
@@ -1,13 +1,45 @@
 package io.avaje.http.api;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
 /*
  Extendable response class, if this class is extended it will be automatically added to the openapi json file.
  The default status code is 400 but can be changed with for example @StatusCode("404")
+
+ To register this response exception do:
+     javalin.exception(BadRequestResponse.class, (badRequestResponse, context) -> {
+        context.json(badRequestResponse.getResponse());
+        StatusCode statusCode = badRequestResponse.getClass().getAnnotation(StatusCode.class);
+        if(statusCode != null)
+            context.status(statusCode.value());
+        else
+            context.status(400);
+    });
  */
 public class BadRequestResponse extends Exception {
   public final String message;
 
   protected BadRequestResponse(String message) {
     this.message = message;
+  }
+
+  public Map<String, Object> getResponse() {
+    Field[] fields = this.getClass().getFields();
+    Map<String, Object> map = new LinkedHashMap<>();
+    for(Field field: fields){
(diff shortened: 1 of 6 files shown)


Tijs-2:
For Kotlin it does work but only when you add something like:
@Throws(BadRequestResponse::class)

rbygrave:
Hi, still on holiday for another 2 weeks. I'll look at this when I am back but looks good !!

Tijs-2:
Hi Rob,

I hope you had a nice holiday!

I was also wondering if you had the time to look at this pull request if it could work or if there are improvements to be made.

Tijs

rbygrave:
Right, sorry for the delay.  

I believe the openapi approach is to add something to the controller method like for example:

```java
@ApiResponse(statusCode=503, MyException.class)
@Post
MyResponse doStuff(MyRequest body) {

}

```
I see some benefit to that approach.  Maybe it is a case of having both as well.  Sorry, I just need to actually get a bit of time to look at this.

Cheers, Rob.

Tijs-2:
No problem, and I think you are correct that that is the most used option.

But personally I like my method better because I like to automate things.
If we can detect that an exception is thrown and that exception has an status code why not add it automatically?
Because I think that a lot of people (me included :) ) forget to add the @ApiResponse when a method can throw an exception.

Although I do admit my method works better in Java then in Kotlin because Java needs to handle the exceptions and Kotlin does not if I am not mistaken.

Kind regards,
Tijs

Tijs-2:
@rbygrave Did you also find some time to look at this merge request?

SentryMan:
This PR has been idle for quite some time, and the `@ApiResponse(statusCode=503, MyException.class)` approach has the advantage of showing the return type schema. So I'm gonna close this.

Issue #51, opened 2021-02-16
Adding endpoint from other libs to openapi documentation


Tijs-2:
Hi Rob,

I was looking into a way to add endpoint from another lib to the openapi documentation of the final service.

For example I have a base project y that provides a general /health or /info endpoint and I use that lib in project z.
The endpoints are made available with the use of your inject project when I start the application, but they are not present in the openapi documentation.

Now I was wondering and trying out but could not achieve this by looking for the "Controller" classes and trying to add them but it is not looking to other inherited projects.

Would something like this be possible in combination with the @ContextModule[depenson=project y] so we know where to look and then add those endpoints? Because it would be nice to add all available endpoints to the final documentation.

Kind regards,
Tijs

rbygrave:
My first thought is something like:  Each module generates the swagger file with a unique name (e.g. health-openapi.json) and a properties resource that points to it's location (/openapi.location.properties with content like location=/health-openapi.json).

Then to "merge" the API's into a single API something could load all the /openapi.location.properties resources and hence then load all the module api json resources and then merge them.

Something like that. Hmmm.

Tijs-2:
Could be an idea bundling the openapi files, I should try it with 2 files. I have done a quick search but did not find a good library but will try again tomorror. Or trying to merge them just as json files...

I have also been checking if I can find the other controllers and maybe add them to the openapi processor from the io.avaje.http.generator.core.BaseProcessor but I cannot search other jar files.

Maybe it cannot be done or it is some knowledge that I lack, could also just be not a good way to go...

Tijs

PS, not really for this issue did you find some time to check my pull request because it would be nice to use it in my application

Tijs-2:
Hi @rbygrave 

I have found a way but it is quite some workand do not know if you are interested in this.
With a method of the org.reflections library I can find all the packages with a BeanContextFactory and within those packages the classes annotated with the right annotations.

Like this:
    List<URL> packages = new ArrayList<>();
    for(URL url: ClasspathHelper.forClassLoader()) {
      Reflections reflection = new Reflections(new ConfigurationBuilder().setUrls(url).setScanners(new ResourcesScanner()));

      if(reflection.getResources(Pattern.compile(".*\\.BeanContextFactory")).size() > 0)
        packages.add(url);
    }

    Reflections reflection = new Reflections(new ConfigurationBuilder().setUrls(packages).setScanners(new SubTypesScanner(), new TypeAnnotationsScanner()));
    try {
[...]

PR #54, opened 2021-03-02
Now we can find controllers in other jar files / merging openapi.json

Tijs-2:
I have tried something, it was a lot of work and I am not sure if it is a good thing but it works.
I now can locate controllers in other jar files and add them to the openapi.json file

So it could be a solution for #51 

But as I said it kind of is a large change on top of your code. It seems to work and I have compared the openapi.json that is created on a few types of parameters / returns types. With and without lists etc.

One thing that is also important to mention you need to add the following to your pom file to keep the parameter names so they can be used for generating the openapi.json file.

        <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <configuration>
                <compilerArgument>-parameters</compilerArgument>
                <source>${maven.compiler.source}</source>
                <target>${maven.compiler.target}</target>
                <parameters>true</parameters>
            </configuration>
        </plugin>

Please let me know if it could be something useful or not...
Tijs

### DIFF
--- a/http-generator-core/pom.xml
+++ b/http-generator-core/pom.xml
@@ -53,7 +53,11 @@
       <artifactId>swagger-annotations</artifactId>
       <version>${swagger.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>org.reflections</groupId>
+      <artifactId>reflections</artifactId>
+      <version>0.9.12</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/http-generator-core/src/main/java/io/avaje/http/generator/core/BaseControllerReader.java
+++ b/http-generator-core/src/main/java/io/avaje/http/generator/core/BaseControllerReader.java
@@ -1,0 +1,52 @@
+package io.avaje.http.generator.core;
+
+import io.avaje.http.api.Path;
+import io.avaje.http.api.Produces;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+
+/*
+ B = Either Class or TypeElement, for Bean
+ I = Either Class or Element, for interfaces
+ M = Either Method or ExecutableElement for interfaceMethods
+ */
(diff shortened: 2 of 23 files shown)


rbygrave:
I think we should be able to "merge" the openapi.json files.  Something like load each of the openapi.json files as resources and then merge the models.

rbygrave:
Apologies this has taken me so long to back on this.

Tijs-2:
No problem, I have also been a bit busy lately. I did look at merging the json files again and found something interesting:
https://github.com/kpramesh2212/openapi-merger-plugin

I did not try it but seems like a nice tool if it works.

Other thing I found was this:
https://stackoverflow.com/questions/39945039/java-merge-json-files

The first one seems the best choice to look at because it is made for openapi files.

Mechite:
Library for this (but it is written in Kotlin) https://github.com/savvasdalkitsis/json-merge
It is a single file, so it will literally be super trivial to convert to Java, I could probably send a gist here for a Java port of it.

Mechite:
https://github.com/flipkart-incubator/zjsonpatch should be used in this case. We can merge until we have one large document, but we should ensure the current projects' openapi.json is the last document to be "appended" so that we end up with its other fields being the ones that take effect.

Trivial example:
```java
    JsonNode merge(JsonNode mainNode, JsonNode updateNode) throws Exception {
        // generate a patch between the main and update JSON objects
        JsonNode patchNode = JsonDiff.asJson(mainNode, updateNode);
        
        // apply the patch to the main JSON object to create a new merged object
        JsonNode mergedNode = JsonPatch.fromJson(patchNode).apply(mainNode);
        
        return mergedNode;
    }
    ```

SentryMan:
this PR is beyond stale

PR #92, opened 2022-09-29
Add more MediaTypes for byte[] responses;

SentryMan:
Since byte arrays will be supported, let's add some common media types used with them;


rbygrave:
Great, thanks.

PR #106, opened 2022-10-20
Enforce Google Java Format with Maven

SentryMan:
.editorconfig files don't really mean much for those that don't use IntelliJ/VSCode. With a plugin, code will be formatted in a standard way no matter the IDE.

Formatting changes only


rbygrave:
> .editorconfig files don't really mean much for those that don't use IntelliJ/VSCode

Really? I have heavy contributors to ebean orm using Eclipse and there has never been a formatting issue

SentryMan:
Formatting is my problem.

PR #118, opened 2023-01-06
Use @Singleton again

SentryMan:
- uses @Component only if Avaje is on the classpath
- uses Jakarta/javax @Singleton depending on classpath
- uses Jakarta @Singleton by default if both jakarta and javax on classpath
- adds a compiler arg to force javax

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,19 @@ public class WidgetController {
   record Widget(int id, String name){};
 }
 ```
-
+## @Singleton
+By default the generated classes will use jakarta.inject.Singleton for wiring. To use Javax Inject, use the compiler arg `-AuseJavax=true` 
+```
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<compilerArgs>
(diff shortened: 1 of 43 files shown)


rbygrave:
Yeah, I don't think this is the best approach.  We ideally have 4 cases to support.

- using avaje-inject Jakarta (ignoring javax)
- using avaje-inject javax (ignoring Jakarta)
- using some other DI via Jakarta 
- using some other DI via javax

As an alternative approach, it can detect avaje-inject and then failing that detect Jakarta or javax.

SentryMan:
I can easily add the avaje detection and use @Component for those cases. But for scenrio 3 and 4, I think defaulting to jakarta unless specified is the way to go. Consider the scenario where javax/jakarta just happens to be on the classpath because of a transitive dependency. How should the processor proceed? A flag makes it simple.

SentryMan:
@rbygrave how about this? 

Will use avaje if it's there. Otherwise will use either jakarta/javax on the classpath. If both are on the classpath, will default to jakarta, unless overridden by the flag.

rbygrave (approved):
Yup, excellent.

PR #120, opened 2023-01-08
Remove Jackson/Custom Serialize OpenAPI json

SentryMan:
So I was mildly inconvenienced when I was using this. It looks like when I try to use the AP with maven provided scope only in a project that doesn't use jackson at all. I get issues.
```
		<dependency>
			<groupId>io.avaje</groupId>
			<artifactId>avaje-inject-generator</artifactId>
			<version>${avaje.inject.version}</version>
			<scope>provided</scope>
		</dependency>
		<dependency>
			<groupId>io.avaje</groupId>
			<artifactId>avaje-jsonb-generator</artifactId>
			<version>${avaje.jsonb.version}</version>
		</dependency>
		<dependency>
			<groupId>io.avaje</groupId>
			<artifactId>avaje-http-javalin-generator</artifactId>
			<version>${avaje.http.version}</version>
			<scope>provided</scope>
		</dependency>
```
I get this error.
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project 
avaje-javalin: Fatal error compiling: java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/ObjectMapper: com.fasterxml.jackson.databind.ObjectMapper -> [Help 1]
```
It seems the jackson dependency isn't transitively pulled while the processor runs. If I add jackson or use the AP through the maven compiler plugin it works fine.

To solve this, I thought about just removing jackson.

### DIFF
--- a/http-generator-core/pom.xml
+++ b/http-generator-core/pom.xml
@@ -13,7 +13,6 @@
 
   <properties>
     <swagger.version>2.0.8</swagger.version>
-    <jackson.version>2.13.4.1</jackson.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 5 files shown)


rbygrave:
Ok, interesting ...

PR #122, opened 2023-01-13
nima modules - make it easy to run / configure / test using Nima

rbygrave:
Work In progress ...


rbygrave:
No it doesn't. 

This is just going to be a relatively fast way to get going using Nima + avaje inject + avaje jsonb ... that includes component testing with the server running on random port.

Still needs a but of work.

The nifty thing this adds is in avaje-nima-test as the plug in for avaje-inject test.  It detects if the test includes the http client and if so starts the Nima webserver on random port for the test. (Avaje jex also has this).

rbygrave:
It does make me think that HttpClientContext should instead be renamed to HttpClient or similar.

rob-bygrave:
Closing as this code has all moved to live at https://github.com/avaje/avaje-nima

PR #123, opened 2023-01-15
Now can define path in @Controller

SentryMan:
Now we can use one annotation to define a controller path instead of using Controller + Path
- Updates Controller annotation to support path
- Updates reader to get the path value from controller (Will prefer Controller annotation over Path annotation to get the path)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,11 +75,9 @@ package org.example.hello;
 
 import io.avaje.http.api.Controller;
 import io.avaje.http.api.Get;
-import io.avaje.http.api.Path;
 import java.util.List;
 
-@Path("/widgets")
-@Controller
+@Controller("/widgets")
 public class WidgetController {
   private final HelloComponent hello;
   public WidgetController(HelloComponent hello) {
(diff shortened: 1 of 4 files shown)


rbygrave:
Yeah noting that there are 3 cases:
1. Controller extends an interface that has `@Path`
2. `@Controller(...) 
3. `@Controller @Path(...)`

It's the controller extending an interface option which means we need to be careful and certainly not just say - always specify the path on the `@Controller("/foo"`

We probably need to make sure the javadoc is better in terms of explaining these 3 options .

rbygrave (approved):

SentryMan:
> 1. Controller extends an interface that has `@Path`
didn't even know this was a thing you could do, I'll add it to the docs.

rbygrave:
Yes, create a java interface that is then shared by both the client and the server.  Put `@Client` on it to have the client generated and create a server Controller that implements the interface. This is a thing that people can do with JAX-RS so really we are just copying JAX-RS with this approach.

Note that this isn't **_"ideal"_** in terms of client execution at scale without Loom Virtual threads ... **_but_** with Virtual Threads this becomes good and as a consequence we effectively will not bother using the `HttpCall`.  

Without virtual threads people would probably look to create a client specific interface and use `HttpCall` (similar to Retrofit).

PR #125, opened 2023-01-16
Fix Annotations when using JsonB

SentryMan:
JsonB has beef with annotations I guess.


http-generator-core/src/main/java/io/avaje/http/generator/core/Util.java
```diff
@@ -30,10 +31,19 @@ public static UType parse(String rawType) {
...
       DeclaredType declaredType = (DeclaredType) typeMirror;
+
       return declaredType.asElement().toString();
     } else {
-      return typeMirror.toString();
+      return trimAnnotations(typeMirror.toString());
+    }
+  }
+
+  static String trimAnnotations(String type) {
```
rbygrave, line 41: ideally we add a unit test for this trimAnnotations() method

rbygrave (approved):

PR #128, opened 2023-01-16
New Client Name

SentryMan:
Updates generators to use the new HTTP Client Name.

Relies on https://github.com/avaje/avaje-http-client/pull/57 and https://github.com/avaje/avaje-http-client/pull/58


rbygrave:
Oh, I think we should still merge this at some point - maybe after a release or 2.

I'm looking to move the avaje-http-client git repo into avaje-http and keep it all together.

SentryMan:
Actually no, generation is busted if you try the new name: try the run.sh on this https://github.com/SentryMan/avaje-javalin-api-example

rbygrave:
Replacing this PR with https://github.com/avaje/avaje-http/pull/133 

.. so may as well close this now.

PR #136, opened 2023-01-17
Simplify useJavax determination in ProcessingContext

rbygrave:
IntelliJ warned that for `javax && !jakarta` the !jakarta part was always true. That leads to this simplification.


SentryMan (approved):
Hmm, yeah now that I think about it this looks good

PR #142, opened 2023-01-19
AutoProvide Validator

SentryMan:
now doesn't need @InjectModules


rbygrave:
Huh, that's pretty groovy !!!  I need to release that avaje-inject version ...

PR #143, opened 2023-01-19
Javax Validator Auto Provide

SentryMan:
javax I guess, should we even merge this?


rbygrave:
I can pull it locally and release it from there without merging it into master.  

I'll comment back into here when I've done that.

SentryMan:
the jakarta one is still not on mvn central

SentryMan:
hmm wait a minute, on my local it doesn't work

SentryMan:
did we get this pushed?

rbygrave:
> did we get this pushed?

No, I've just done it now ... so version 2.9 is on it's way to maven central now.

PR #145, opened 2023-01-20
Add module-info to Hibernate validator (Jakarta)

rbygrave:
Also remove @InjectModule as we just use it as a avaje-inject plugin (so it should not be both)


rbygrave:
(version 3.1)

PR #146, opened 2023-01-20
Update validator version

SentryMan:
also fix meta-inf directory


rbygrave:
I'll merge this in.  Note that 3.1 never made it to maven central so I'll put that version back I think.

rbygrave:
Looking to release 3.1-RC1 ... but maven central is very unhappy at the moment.  Will update here if it is successful.

PR #148, opened 2023-01-25
Fix such that Class Roles are added to Method Roles

SentryMan:
Currently getting OperationUnsupportedException

--------
Edit by Rob

This change is such that the roles for a given method are the roles defined on the method _**PLUS** the roles defined on the controller class / type_.


SentryMan:
09e3c79f19197f939573d1e8a5e1e7253d8423b1
9544536bba136d90985b9696de5ab11d203a9f39
these two commits have broke the client tests

PR #149, opened 2023-01-28
Lengthen Inner Type Short Names

SentryMan:
Change JsonB Generation to use expanded short names for inner types. Fixes cases where inner types have conflicting short names.
Now can have a controller like this.
```
  @Post("/outer1/state")
  List<Outer1.State> testInnerClass(Outer1.State state) {

    return List.of(state);
  }

  @Get("/outer2/state")
  Outer2.State testInnerClass2() {

    return new Outer2.State();
  }
```

The generated route will now have the JSON type adapters look like this:
```
  private final JsonType<Outer1.State> outer1$StateJsonType;
  private final JsonType<Outer2.State> outer2$StateJsonType;
```

Related to https://github.com/avaje/avaje-jsonb/pull/55


http-generator-core/src/main/java/io/avaje/http/generator/core/UType.java
```diff
@@ -202,12 +202,36 @@ public Set<String> importTypes() {
       Set<String> set = new LinkedHashSet<>();
       for (String type : allTypes) {
         if (!type.startsWith("java.lang.") && type.indexOf('.') > -1) {
-          set.add(type.replace("[]", ""));
+          if (type.startsWith("java")) set.add(type.replace("[]", ""));
+          else set.add(innerTypesImport(type).replace("[]", ""));
```
rbygrave, line 206: note to self - add blocks

PR #153, opened 2023-02-06
http version in test parent pom

SentryMan:
Doesn't it get tedious modifying every test pom for every version?


rbygrave:
Yes and no. It's just a single global search n replace, not much difference from changing it in one place as changing it in many ... but I'll still merge this in :)

PR #155, opened 2023-02-10
Use Prisms instead of using annotations directly

SentryMan:
Unfortunately, I still had to use the shade plugin to get it to work on a modular project.

- use prisms instead of annotation
- use antrun plugin to remove shaded deps from module-info
- use moditect plugin to add modules-info into shaded jars
- adds support for jakarta Email/Size/etc annotations
- now reads all tag annotations from an element

Relies on https://github.com/avaje/avaje-prisms/pull/7


SentryMan:
I think I'm good with this

Issue #154, opened 2023-02-09
Support Javalin async request handling


LoonyRules:
I would like to begin a conversation regarding planning a possible implementation to support Async Request Handling in the `avaje-http-javalin-generator` as right now, there is no support for such.

If you attempt to do:
```java
@Get("/async")
public CompletableFuture<MyCustomPojo> asyncEndpoint() {
  return CompletableFuture.supplyAsync(() -> {
    return new MyCustomPojo("hello", "world");
  });
}
```
then you will be met by a compile time error stating that CompletableFuture is not an allowed return type.

If you try:
```java
@Get("/async")
public void asyncEndpoint(Context ctx) {
  ctx.future(() -> {
    return CompletableFuture.supplyAsync(() -> {
      return new MyCustomPojo("hello", "world");
    }).thenAccept(ctx::result);
  });
}
```
then Avaje will write the response and send it off, before the future is even completed.

Now, this does not stop you from having Async Request Handling if you use avaje-http with Javalin, however, you will have to do all of that in Javalin land rather than avaje-http. Great for a temporary solution, but it'd be nice to have support in avaje-http itself.

----
There are a few ways that Javalin supports async request handling.

### 1. Context.future(Supplier<CompletableFuture<?>>)
**Javalin:**
```java
javalin.get("/async", ctx -> {
  ctx.future(() -> {
    return CompletableFuture.supplyAsync(() -> {
      return new MyCustomPojo("hello", "world");
    }).thenAccept(ctx::result);
  });
});
```

**Avaje Proposal:**
```java
@Get("/async")
public CompletableFuture<MyCustomPojo> asyncEndpoint() {
  return CompletableFuture.supplyAsync(() -> {
    // Pretend this is an expensive operation (eg: IO)
    return new MyCustomPojo("hello", "world");
  });
}
```

Although this would support the server defined Async Timeout that Javalin has, it would not be an overridable option for this method.

### 2. Context.async(ExecutorService executorService, long timeout, Runnable runnable, ThrowingRunnable<Exception> throwingRunnable)
**Javalin:**
```java
javalin.get("/async", ctx -> {
  ctx.async(
    // Obviously, never do this.
    Executors.newCachedThreadPool(),
    // Timeout in Milliseconds.
    10_000L,
    // On Timeout
    () -> {
      // Do something when the request handling times out.
      // For example, write a timeout response, or throw a 
      // timeout exception for handling elsewhere.
    },
    // Request Handling Runnable
    () -> {
[...]

PR #156, opened 2023-02-11
Support Javalin async request handling [Context#future(...)]

LoonyRules:
This PR adds support for async request handling for Javalin users, using the `Context#future(...)` method as per the discussion in #154


SentryMan:
Can you also put an async path on the Javalin-jsonb test package?

LoonyRules:
> Can you also put an async path on the javalin-jsonb test controller?

@SentryMan I actually cannot do that as the generator creates 2 global variable fields with the same signature instead of just knowing what fields have been created, and use it, if it detects a same signature would've been created. This is something I mentioned to you on discord a couple of weeks ago but looks like we forgot to make a ticket.

Here's the code change:
```java
  @Hidden
  @Get
  List<HelloDto> getAll() {
    return myService.findAll();
  }

  // Added this method
  @Get("/async")
  CompletableFuture<List<HelloDto>> getAllAsync() {
    return CompletableFuture.supplyAsync(() -> {
      // Simulate a delay as if an actual IO operation is being executed.
      // This also helps ensure that we aren't just getting lucky with timings.
      try {
        Thread.sleep(10L);
      } catch (InterruptedException e) {
        throw new RuntimeException(e);
      }

      return myService.findAll();
    }, Executors.newSingleThreadExecutor()); // Example of how to use a custom executor.
  }
```
both of these methods would have the same `JsonType` signature and variable names so that's what causes the error:
```java
  private final JsonType<List<HelloDto>> listHelloDtoJsonType;
  private final JsonType<List<HelloDto>> listHelloDtoJsonType;
  private final JsonType<HelloDto> helloDtoJsonType;

  public HelloController$Route(HelloController controller, Validator validator, Jsonb jsonB) {
    this.controller = controller;
    this.validator = validator;
    this.listHelloDtoJsonType = jsonB.type(HelloDto.class).list(); // GET /hello (getAll)
    this.listHelloDtoJsonType = jsonB.type(HelloDto.class).list(); // GET /hello (getAllAsync)
    this.helloDtoJsonType = jsonB.type(HelloDto.class);
  }
``` 
`variable listHelloDtoJsonType is already defined in class org.example.myapp.web.HelloController$Route`

http-generator-javalin/src/main/java/io/avaje/http/generator/javalin/ControllerWriter.java
```diff
@@ -87,7 +87,13 @@ private void writeClassStart() {
       writer.append("  private final Validator validator;").eol();
     }
 
-    for (final UType type : jsonTypes.values()) {
+    for (UType type : jsonTypes.values()) {
+      // Support for CompletableFuture's.
+      if (type.isGeneric() && type.mainType().equals("java.util.concurrent.CompletableFuture")) {
```
SentryMan, line 92: you need to add a check here so that the raw type of the future doesn't get written twice.

SentryMan: I guess either modify the `jsontypes` map, or create a set that tracks what future types have been written. Other non- types do not get duplicates.

SentryMan: I'm thinking of doing something like this. 
```
      for (UType type : jsonTypes.values()) {
        // Support for CompletableFuture's.
        if ("java.util.concurrent.CompletableFuture".equals(type.mainType())) {
          type = type.paramRaw();
          if (jsonTypes.containsKey(type.full()) || !futureSet.add(type.full())) {
            continue;
          }
        }
```

SentryMan: You'll probably need to take out the` throw IllegalStateException` from the JsonbUtil and put it into the first for loop in the ControllerWriter.

LoonyRules: Now sure how I didn't notice the `jsonTypes` map and that it could be used to fix the issue I came across.

I've moved the `CompletableFuture` generic validation over to the `ControllerMethodWriter` for Javalin as it should then provide better context (controller + method) in regards to _what_ is wrongly defined.

Async tests in `test-javalin-json` have also been added.

SentryMan (approved):
LGTM

rbygrave (approved):

PR #159, opened 2023-02-14
Fix future jsonb generation

### LINKED ISSUE #158: Javalin Controller not writing JsonType variable assignment generated when using a CompletableFuture
Knew there'd be something we missed when adding this functionality.

**What happened?**
If you define a controller with a single endpoint that has a `CompletableFuture<List<String>>` as a return type, then the generator does not write the `this.listStringJsonType = ...;` line of code. 

If you had another endpoint in the same controller that did return `List<String>` as a return type, then you wouldn't notice the issue, this is how our HelloControllerTests work and why they didn't pick up on the issue.

**What is expected?**
The `this.listStringJsonType = ...;` line generation should be written at least only once, and not doing some hackery to stop the line being generated twice.

----
Both @SentryMan and I have a similar approach for fixing this, so we'll work something out so a v1.27 can be release ASAP. A temporary work around is outlined in the second paragraph of `What Happened?`.

SentryMan:
- adds a set to track future adapters
- adds a method to UType to detect if the type is a completable future

fixes #158

### DIFF
--- a/http-generator-core/pom.xml
+++ b/http-generator-core/pom.xml
@@ -75,7 +75,6 @@
       <groupId>io.swagger.core.v3</groupId>
       <artifactId>swagger-annotations</artifactId>
       <version>${swagger.version}</version>
-      <optional>true</optional>
       <scope>provided</scope>
     </dependency>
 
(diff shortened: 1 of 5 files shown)


SentryMan:
I like #160 better

PR #163, opened 2023-02-17
support `@SecurityScheme`

kevin70:
Hello.
I added support for `@SecurityScheme` and `@SecurityRequirement`.

Use the following method to automatically convert security authentication to the corresponding OpenAPI.
e.g.
```
@Controller
@Path("/security")
class SecurityController {

  @Get("/first")
  @SecurityRequirement(name = "JWT")
  String first() {
    return "simple";
  }

  @Get("/second")
  @SecurityRoles
  String second() {
    return "simple";
  }
}
```

There seems to be a lot of source files that I changed this time. Please help me review them if you have time. If there is any problem, I will modify it in time. Thank you.


SentryMan:
ok so clearly my custom openAPI serializer doesn't cut it anymore. Instead of pulling all of swagger core, can you only add jackson databind and do something like [how we had it before](https://github.com/SentryMan/avaje-http/blob/8d85eca4b20495de988a9d3a3a158f8f53ae3232/http-generator-core/src/main/java/io/avaje/http/generator/core/openapi/DocContext.java#L179-L200)?

kevin70:
swagger code has a large number of optimization implementations for serialization. Using swagger-code can better maintain compatibility with the OpenAPI specification. This is helpful for generating better OpenAPI documents, so I hope to introduce swigger-code to better help us solve this problem.

SentryMan:
> swagger code has a large number of optimization implementations for serialization. 

If a regular ObectMapper can serialize correctly, do we need all of the optimizations? I guess I'm hesitating to add all the transitive dependencies to the shaded jar (jakarta.xml.bind, commons-lang, snakeyaml, etc.) if we don't actually need them.

kevin70:
I removed the swagger core dependency and replaced it with jackson. The generated OpenAPI document looks good at present.

SentryMan (approved):
LGTM

SentryMan:
Wait a second, we do have an issue it seems, when I use avaje-http(your changes installed to my local maven) as a provided dependency I get this error.
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Failed on call to `getDeclaredMethods()` on class `io.swagger.v3.oas.models.OpenAPI`, problem: (java.lang.NoClassDefFoundError) io/swagger/v3/oas/models/security/SecurityRequirement
```
We'll have to figure out why it can't find the model class

SentryMan:
I get this error with or without the `@SecurityRequirement` annotation, I also tried reverting to use swagger core deserialization and it still doesn't work. To reproduce, mvn clean install your changes, then clone https://github.com/SentryMan/avaje-javalin-api-example and bump the version in the POM.

SentryMan:
also it seems using the ObjectMapper by default outputs invalid openAPI when validated by https://apitools.dev/swagger-parser/online/

kevin70:
> also it seems using the ObjectMapper by default outputs invalid openAPI when validated by https://apitools.dev/swagger-parser/online/

Yes, thank you for your repair. I ignored this problem at the beginning.

SentryMan:
we can close this since the changes will be added in #164 (when it gets reviewed)

PR #167, opened 2023-02-27
Support Multi-Value QueryParam/Headers/Cookies

SentryMan:
Got a request from discord to add this.

- Creates a new Collection Type Handler specifically for those param types
- adds set, list, and a collection version of `withDefault` method to Param Converter
- adds new Platform Adapter method to handle multi-value parameters
- adds new Platform Adapter method to handle map parameters
- Can do `Map<String, List<String>>` for certain params.

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
@@ -28,8 +28,8 @@ public final class PathTypeConversion {
     return value != null ? value : defaultValue;
   }
 
-  public static List<String> withDefault(List<String> value, String defaultValue) {
-    return value != null && !value.isEmpty() ? value : List.of(defaultValue);
+  public static List<String> withDefault(List<String> value, List<String> defaultValue) {
+    return value != null && !value.isEmpty() ? value : defaultValue;
   }
 
   /**
--- a/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
@@ -3,6 +3,7 @@ package io.avaje.http.api;
 import java.math.BigDecimal;
 import java.time.*;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
--- a/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
@@ -2,7 +2,11 @@ package io.avaje.http.api;
 
 import java.math.BigDecimal;
 import java.time.*;
+import java.util.List;
+import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * Helper type conversion methods.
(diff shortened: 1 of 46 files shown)


http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
```diff
@@ -22,6 +28,10 @@ public static String withDefault(String value, String defaultValue) {
     return value != null ? value : defaultValue;
   }
 
+  public static List<String> withDefault(List<String> value, String defaultValue) {
```
LoonyRules, line 31: Just allow us to define a `List<String> defaultValue` parameter instead? Some defaults may include multiple entries.

http-generator-core/src/main/java/io/avaje/http/generator/core/ElementReader.java
```diff
@@ -56,6 +70,46 @@ public class ElementReader {
...
+      final var mainTypeEnum =
+          typeOp
+              .flatMap(t -> Optional.ofNullable(ctx.typeElement(t.mainType())))
+              .map(TypeElement::getKind)
+              .filter(ElementKind.ENUM::equals)
+              .isPresent();
+
+      final var isCollection =
+          typeOp
+              .filter(t -> t.isGeneric() && !t.mainType().startsWith("java.util.Map"))
```
LoonyRules, line 89: `Optional<>` params are gonna haunt you with this.

http-generator-core/src/main/java/io/avaje/http/generator/core/ElementReader.java
```diff
@@ -56,6 +70,46 @@ public class ElementReader {
...
+              .isPresent();
+
+      if (mainTypeEnum) {
+        return TypeMap.enumParamHandler(type);
+      } else if (isCollection) {
+        this.isParamCollection = true;
+        if (paramType == ParamType.FORMPARAM) {
+          throw new IllegalStateException("You can't have a single Form Parameter be a list");
```
LoonyRules, line 97: Code example for what this is trying to catch? Doesn't sound right to me?

http-generator-core/src/main/java/io/avaje/http/generator/core/ElementReader.java
LoonyRules: Also, what if my BeanParam has a Map<String, String> field in it? Retrofit2 supports all of this stuff and can be deserialized by general json deserialization as any "BeanParam" would be treated as a Body request (GET/POST/PATCH/PUT). Could we just do the same?

SentryMan: Oh we already support maps

http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
SentryMan: ok

http-generator-core/src/main/java/io/avaje/http/generator/core/ElementReader.java
SentryMan: > `Optional<>` params are gonna haunt you with this.

Well, it's a good thing we don't support `Optional<>` then.

SentryMan:
And I'm done, @rbygrave  plz review when you get time.

SentryMan:
Split HttpClient changes into it's own PR #169

PR #168, opened 2023-02-28
Add Request Timeout Annotation

SentryMan:
Now can override the global timeout on the generated client interfaces


rbygrave:
Another nice addition !!

PR #175, opened 2023-03-03
fix client processor overriding OpenAPI

### LINKED ISSUE #174: The 1.28 version in gradle cannot correctly generate openapi
```
version: 1.28
```

Generated:
```
{
	"openapi" : "3.0.1",
	"info" : {
		"title" : "",
		"version" : ""
	},
	"paths" : {
		
	},
	"components" : {
		"schemas" : {
			
		}
	}
}
```
**I return the version to 1.27 and can generate openapi normally**.

The new `DocContext` cannot work normally under the Gradle. `AbstractProcessor.process` will be executed repeatedly. The subsequent execution overwrites the previous data. I'm sure that the `AbstractProcessor.init` function will only be executed once in the Gradle. Maybe we should initialize the `DocContext` in the init function.

SentryMan:
I guess the Client Processor would overwrite the DocContext in the processing context

Fixes #174


kevin70:
Yes, you are right. I can work normally after deleting the `avaje-http-client-generator` dependency.

PR #177, opened 2023-03-09
OpenAPI documents support enumeration types

kevin70:
微信截图_20230309104611

I tried to add support for enumeration types in the OpenAPI document, but the current JSON cannot handle it correctly. Whether to consider introducing swagger core to support document processing? I feel this is necessary.


SentryMan:
> Whether to consider introducing swagger core to support document processing? I feel this is necessary.

remember we tried that before and it also didn't create the json correctly

SentryMan:
In line 84 of the OpenApiSerializer, replace the existing line with this.

sb.append(field.getName().replace("_enum", "enum"));

rbygrave:
Is this PR completed now?  It looks finished and ready to me.

kevin70:
Yes, it's finished. Thanks.

SentryMan:
LGTM

PR #181, opened 2023-03-12
Improve Javadoc only

rbygrave:


SentryMan:
LGTM, though when you deploy, can you make it an RC instead of a full release, please? I've got a bunch of stuff in the works

rbygrave:
I don't mind it being `1.31` ... all good to have more stuff coming shortly.

PR #185, opened 2023-03-15
Single ThreadLocal Context

SentryMan:
You already knew this was coming


SentryMan:
Ah hang on it's not working with the client processor

SentryMan:
Okay, I'm good with this. @rbygrave please review and deploy when you get time.

PR #189, opened 2023-03-20
Add `requires java.sql` to generators

### LINKED ISSUE #188: IntelliJ - IllegalAccessError: class io.avaje.http.generator.core.openapi.KnownTypes ... does not read module java.sql
When compiling with IntelliJ (we don't get this error with maven) we can get the error below:

```
java: java.lang.IllegalAccessError: class io.avaje.http.generator.core.openapi.KnownTypes (in module io.avaje.http.nima.generator) cannot access class java.sql.Date (in module java.sql) because module io.avaje.http.nima.generator does not read module java.sql

```

The generator modules use the generator-core and with avaje-prisms and shade we don't need those module-info requires in generator-core EXCEPT it seems for `java.sql` which is used as part of the openapi generation to support the `java.sql.Date` type.

The fix for this issue is to include `requires java.sql` in the generator modules that use generator-core.

SentryMan:
Fixes #188

### DIFF
--- a/http-generator-client/src/main/java/module-info.java
+++ b/http-generator-client/src/main/java/module-info.java
@@ -3,6 +3,7 @@ module io.avaje.http.client.generator {
   provides javax.annotation.processing.Processor with io.avaje.http.generator.client.ClientProcessor;
 
   requires java.compiler;
+  requires java.sql;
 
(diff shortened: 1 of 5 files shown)


SentryMan:
It seems the race does not always go to the swift.

PR #191, opened 2023-03-21
Nima JsonB Stream

SentryMan:
Now for smaller payloads, helidon nima now will not duplicate the byte array (the makes jsonb the first json serializer to not auto copy the buffer for short requests)


SentryMan:
hmm, forgot that nima only compiles to 19

http-api/src/main/java/io/avaje/http/api/NimaJsonbStream.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public void write(int b) throws IOException {}
+
+  @Override
+  public void write(byte[] b, int off, int len) throws IOException {
+    if (chunked) {
+      outputStream.write(b, off, len);
+    } else if (firstBuffer == null) {
+      firstBuffer = b;
```
rbygrave, line 29: I think there is a subtle issue here in that the buffer will get reused / overwritten (if there are 2 or more "buffer writes" / if content exceeds 4k).

That is, I don't think we can keep a reference to b, off, len ... in the form of firstBuffer etc. We'd need to make a defensive copy as we don't know if this was the "first and last write" or "just the first or many writes".

SentryMan:
Now I use the new Jsonb stuff

SentryMan:
ah woops, forgot to bump the jsonb versions

PR #194, opened 2023-04-02
[http client] Use more generic Type rather than ParameterizedType

SentryMan:
Hmm, I'm thinking it would be good if we loosen the restriction from only parameterized types.


http-client/src/main/java/io/avaje/http/client/HttpException.java
```diff
@@ -75,15 +75,15 @@ public HttpException(int statusCode, Throwable cause) {
   }
 
   HttpException(HttpResponse<?> httpResponse, DHttpClientContext context) {
-    super();
+    super("Http Call failed with status: " + httpResponse.statusCode());
```
rbygrave, line 78: I think I'll change that to "Http call ..." using lower case "c" for "call" as we have `HttpCall` as a type

PR #195, opened 2023-04-02
Add io.avaje.http.api @Valid Annotation to support custom validators

SentryMan:
So I find myself writing my own custom validators rather than using hibernate, would like this so I can exclude any Jakarta annotation dependency.


tests/test-javalin-jsonb/src/main/java/org/example/myapp/web/GetBeanForm.java
```diff
@@ -30,6 +32,12 @@ public class GetBeanForm {
 
   @QueryParam private Set<ServerType> type;
 
+  @Ignore @IgnoreBeanField private String ignored;
```
rbygrave, line 35: I'm not 100% convinced on this / `@IgnoreBeanField`. My gut says the field should be `transient` and that jsonb and avaje-http should honour transient fields as being ignored.

SentryMan: you know, I always forget that is a valid identifier

SentryMan: I see that transient is used for serialization only though

rbygrave: > I see that transient is used for serialization only though

The transient modifier isn't _just_ used by Java serialisation.  JPA / ORMs similarly honour transient fields/properties and avaje-jsonb _should_ honour transient fields in that jsonb should treat transient fields as `@Json.Ignore` by default.

My thought is that "beans" with transient fields should by default be ignored in terms of query parameters or form parameters.

SentryMan: I meant it's used to say a field should not be serialized, can't seem to find where it says that it also means a field cannot be deserialized.

SentryMan:
gonna split the bean ignore into it's own thing

PR #196, opened 2023-04-03
Use Context#bodyStreamAsClass(...)

Mechite:
See #193 for details.


SentryMan:
On my local this works

PR #197, opened 2023-04-03
Add @Ignore for ignoring a BeanParam field (or query param field)

SentryMan:
Also adds an @Ignore annotation for bean parameters. (have a case where I use the bean param object constructors to initialize other useful fields that are not expected parameters).

It's like a `Jsonb.Ignore(serialization=true)` but for http.

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/IgnoreBeanField.java
+++ b/http-api/src/main/java/io/avaje/http/api/IgnoreBeanField.java
@@ -1,0 +1,12 @@
+package io.avaje.http.api;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/** Mark a field on a BeanParam/FormParam class as not a request parameter of any kind */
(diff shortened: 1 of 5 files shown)


http-api/src/main/java/io/avaje/http/api/IgnoreBeanField.java
```diff
@@ -0,0 +1,12 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/** Mark a field on a BeanParam/FormParam class as not a request parameter of any kind */
+@Target(FIELD)
+@Retention(SOURCE)
+public @interface IgnoreBeanField {}
```
rbygrave, line 12: I think this should just be called `Ignore`.

I realise that is a kind of name clash with `@Json.Ignore` but I think the changes of a "form bean" or "query params bean" also being a `@Json` type is relatively low.

In the case of a clash people can use `@Json.Ignore @Ignore private String ignored`

rbygrave (commented):
Do we want a test for this - confirm the input is ignored?

PR #204, opened 2023-04-16
Add HttpRequest Feature

SentryMan:


http-api/src/main/java/io/avaje/http/api/context/ThreadLocalResolver.java
```diff
@@ -0,0 +1,54 @@
+package io.avaje.http.api.context;
+
+import java.util.Optional;
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+
+public final class ThreadLocalResolver implements RequestContextResolver {
```
LoonyRules, line 7: This is a little nit-picky but I feel like the impl class name should be `ThreadLocalRequestContextResolver`, if you just saw `new ThreadLocalResolver()` used in some code to set the global resolver, then it’s not really obvious that the class is a RequestContextResolver implementation.

http-api/src/test/java/io/avaje/http/api/context/ThreadLocalResolverTest.java
```diff
@@ -0,0 +1,48 @@
+package io.avaje.http.api.context;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
```
LoonyRules, line 6: A test to verify that it all works with a CompletableFuture being the endpoint’s method return type? Seeing as that code in the generated controller sources is different to everything else.

rob-bygrave:
How does this API work for servers that use request + response rather than the single context parameter?  `resolver.supplyWith(ctx,` ... is oriented to servers like Javalin that have a single context parameter.

SentryMan:
Hmm, so it is

SentryMan:
I'll modify the API to handle it

SentryMan:
I added a ServerContext so that both can be retrieved

SentryMan:
I'm good with most of this, but not sure about adding the inject plugin given that whole module issue. Do you think I should keep or remove it?

rob-bygrave:
Lets keep it and see how things evolve over the next week.

PR #206, opened 2023-04-18
Don't export SPI package

SentryMan:
We don't need it to service load.

EDIT: Ahh, but hmm, if the module thing doesn't/won't get fixed we will need to export it.


rob-bygrave:
> but hmm, if the module thing doesn't/won't get fixed we will need to export it.

Yes exactly. I'm honestly not hopeful of getting the fix in the JDK. I think it will take someone like Spring jumping on this issue to get it changed.

SentryMan:
Yeah, this isn't gonna happen

Issue #208, opened 2023-04-20
`ServerContext`'s request() and response() methods are not accessible.


LoonyRules:
The `ServerContext` class has 2 methods:
- `<T> T request() { ... }`
- `<T> T response() { ... }`
but they are not accessible outside of its package, meaning the end user cannot use them. These need to be public methods instead.

rob-bygrave:
> not accessible outside of its package

This could be a nice little change to do.  Do you want to submit a PR?

PR #209, opened 2023-04-20
Fix BeanParam imports

### LINKED ISSUE #207: Missing import for Record BeanParam's that have a collection as a parameter
In `avaje-http` v1.36 the following endpoint definition:
```java
@Get("myEndpoint")
public void myEndpoint(@BeanParam MyRecord myRecord) {
  ....
}
```
with the following record definition:
```java
public record MyRecord(@QueryParam("myEnum") Set<MyEnum> myEnums) { }
```
generates the following code:
```java
ApiBuilder.get("myEndpoint", ctx -> {
  ctx.status(200);
  var myRecord = new MyRecord(
    set(qp -> (MyEnum) asEnum(MyEnum.class, qp), ctx.queryParams("myEnum")
  );

  ...
});
```
but fails to import the `MyEnum.class`, so the generated code is invalid resulting in an error at compile time.

SentryMan:
also fix that server context methods aren't public

fixes #207 and #208

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/context/ServerContext.java
+++ b/http-api/src/main/java/io/avaje/http/api/context/ServerContext.java
@@ -19,7 +19,7 @@ public final class ServerContext {
    * @return The request
    */
   @SuppressWarnings("unchecked")
-  <T> T request() {
+  public <T> T request() {
     return (T) request;
   }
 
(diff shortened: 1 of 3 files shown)


SentryMan:
wow, that took longer than I thought it would

PR #212, opened 2023-04-27
Fix Hard Inject Dependency (use cyclical plugin module)

SentryMan:


rob-bygrave:
I'm not convinced we are going to do this yet.  avaje-inject already has the PR merged that does not use ServiceLoader for Plugins (I merged that in on the 18th and 9.1-RC1 was released).

So this is not the change / PR that brings avaje-http in line with  avaje-inject 9.1-RC1.

SentryMan:
> I'm not convinced we are going to do this yet. avaje-inject already has the PR merged that does not use ServiceLoader for Plugins (I merged that in on the 18th and 9.1-RC1 was released).
> 
> So this is not the change / PR that brings avaje-http in line with avaje-inject 9.1-RC1.

This is but one option, when you make a hard decision for or against we can close/merge as needed.

rob-bygrave:
Alright, noting that people reading this PR should understand that this solves 1 problem but as it stands creates another bigger problem (unless this PR is enhanced further by adding in the cyclical dependency ... and then we have to deal with the build/release issue related to that etc and that gets messy.  Doable Yes, but messy). 

So this PR currently isn't yet an "Acceptable Done Done" solution in my view without dealing with that maven dependency issue.

SentryMan:
> adding in the cyclical dependency

It's already there?

SentryMan:
> The side effect of the maven cyclical dependency is that we need to maintain the `version - 1` version in the plugin pom.

The plugins don't really change that much all things considered.

PR #213, opened 2023-04-27
Fix Hard Dependency on Inject (Rely on custom serviceloader)

SentryMan:


rbygrave:
Closing as we merged the PR that moves the plugin into a separate artefact

PR #218, opened 2023-05-29
Change how Generated Clients are registered.

### LINKED ISSUE #217: Make registering http client interfaces easier for modular applications
ServiceLoader is great, but on a modular application, we need to manually write a `provides` statement for every client interface. This is kind of a hassle, so I'm thinking we could register the providers more like Jsonb's `GeneratedComponent`. In this way, we only ever need to specify a single `provides` for all the generated Client interfaces.

SentryMan:
- Make `HttpApiProvider` a functional interface
- now generates a `GeneratedComponent` class that adds `HttpApiProvider`s to ```DHttpApi```
- no longer generates a provider nested class. (the generated client constructor is used as an HttpApiProvider)
- client processor test for easier debugging of issues 
- can reflectively retrieve implementations using `HttpClient` as well as `HttpClientContext`. 

fixes #217 

this is how the generated component looks

```java
@Generated("avaje-client-generator")
public class GeneratedHttpComponent implements HttpClient.GeneratedComponent {
  @Override
  public void register(Map<Class<?>, HttpApiProvider<?>> providerMap) {
    providerMap.put(ApiClient.class, ApiClientHttpClient::new);
  }
}
```

### DIFF
--- a/http-client/src/main/java/io/avaje/http/client/HttpApiProvider.java
+++ b/http-client/src/main/java/io/avaje/http/client/HttpApiProvider.java
@@ -1,20 +1,20 @@
 package io.avaje.http.client;
 
+import java.util.Map;
+
 /**
  * Provides http client implementations for an interface.
  *
  * @param <T> The interface type
  */
+@FunctionalInterface
 public interface HttpApiProvider<T> {
 
-  /**
-   * Return the interface type this API implements.
-   */
-  Class<T> type();
+  /** Return the interface type this API implements. */
+  default Class<T> type() {
+    throw new UnsupportedOperationException();
+  }
 
-  /**
-   * Return the provided implementation of the API.
-   */
+  /** Return the provided implementation of the API. */
   T provide(HttpClient client);
-
 }
--- a/http-api/src/main/java/io/avaje/http/api/InstrumentServerContext.java
+++ b/http-api/src/main/java/io/avaje/http/api/InstrumentServerContext.java
@@ -18,7 +18,7 @@ import java.lang.annotation.Target;
  * @Get
  * @InstrumentServerContext
  * void helloWorld(long id) {
- *  Server resolver.currentRequest()
+ *  Context = resolver.currentRequest()
  *   ...
  * }
  *
--- a/http-api/src/main/java/io/avaje/http/api/spi/MetaData.java
+++ b/http-api/src/main/java/io/avaje/http/api/spi/MetaData.java
@@ -1,0 +1,18 @@
+package io.avaje.http.api.spi;
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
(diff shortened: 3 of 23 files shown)


http-client/src/main/java/io/avaje/http/client/HttpApiProvider.java
```diff
@@ -1,20 +1,20 @@
...
+@FunctionalInterface
 public interface HttpApiProvider<T> {
 
-  /**
-   * Return the interface type this API implements.
-   */
-  Class<T> type();
+  /** Return the interface type this API implements. */
+  default Class<T> type() {
+    throw new UnsupportedOperationException();
```
rob-bygrave, line 15: This does not look right to me ... hmm

SentryMan: basically, type should not be used

SentryMan: I'm keeping it only for the sake of backwards compatibility

rob-bygrave:
It does not look like this supports "partial compilation" yet?  That is, avaje-jsonb has `io.avaje.jsonb.generator.ComponentReader` ... that reads an existing GeneratedComponent and if exists reads the meta data (in this case factories).  Currently it looks like with a partial compile we can effectively lose factories (when they are not included in the compilation).

SentryMan:
oh so that's what that does

SentryMan:
do we also need a metadata annotation to make this work as well? I never actually understood their purpose in Jsonb.

rob-bygrave:
> metadata annotation to make this work as well?

Yes. 

If there was no such thing as "partial compile" then we would not need the meta data annotations that we use for json and inject.  The main purpose of those "meta annotations" is to support partial compilation where (typically via IDE) a compilation occurs that say only includes a new http client and does not include some http clients that already exist.  For this compile, the processing reads the "meta annotations" (looks up meta-inf services file, finds that element, if exists reads those meta annotations) ... that this tells it that when it generates the GeneratedComponent to include those http clients that already exist.

For jsonb we previously didn't need this because in theory we generated a meta-inf services entry per http client [but it looks to me like we have a bug there in that technically we should APPEND an entry in there for each http client and it looks like it OVERRIDES instead ... so it looks like we have a similar'ish bug in there today]

http-client/src/main/java/io/avaje/http/client/HttpApiProvider.java
SentryMan: when 2.x rolls around I want to take it out actually

SentryMan:
I no longer get the `this class will not be subject to annotation processing warning` so we should be good?

rbygrave:
Yes, it looks all good as I see it - I'll merge it in. 

I should be able to manually test partial compile via IntelliJ (I'll likely manually bump up some logging while I do that etc). Opps, dinner first ...

PR #222, opened 2023-06-02
Fix Enum Path Parsing

### LINKED ISSUE #221: javalin http server generator not parsing path enums
Hello 😄 

Having issues putting enums in the path. I am pretty sure I am not doing anything wrong, according to the [type conversions section in the docs.](https://avaje.io/http/#type-conversions)

Here is an example of the endpoint, where Preference is an enum defined in another subproject.

```java
@Controller("/profile/{playerId}/preferences")
public class PreferenceController {

  @Get("/{preference}")
  public PreferenceOption getPreference(UUID playerId, Preference preference) { .. }
```

But it generates this code,
```java
    ApiBuilder.get("/profile/{playerId}/preferences/{preference}", ctx -> {
      ctx.status(200);
      var playerId = asUUID(ctx.pathParam("playerId"));
      var preference = java.util.Objects.toString(ctx.pathParam("preference")); // not parsed into the enum
      var result = controller.getPreference(playerId, preference); // error here, expecting enum
      ctx.json(result);
    });
```

The error being:
```error
error: incompatible types: String cannot be converted to Preference
      var result = controller.getPreference(playerId, preference);
                                                      ^`
```

Sorry if I misunderstood something, but I think it is theoretically correct.

I am using
- io.javalin:javalin:5.5.0
- io.avaje:avaje-http-api:1.40
- io.avaje:avaje-inject-generator:9.2
- io.avaje:avaje-http-javalin-generator:1.40

Thank you 😄

SentryMan:
fixes #221

### DIFF
--- a/http-generator-core/src/main/java/io/avaje/http/generator/core/TypeMap.java
+++ b/http-generator-core/src/main/java/io/avaje/http/generator/core/TypeMap.java
@@ -310,12 +310,12 @@ class TypeMap {
 
     @Override
     public String toMethod() {
-      return "(" + type.shortType() + ") asEnum(" + type.shortType() + ".class,";
+      return "(" + type.shortType() + ") asEnum(" + type.shortType() + ".class, ";
     }
 
     @Override
     public String asMethod() {
-      return "asEnum(" + type.shortType() + ".class,";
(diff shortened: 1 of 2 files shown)


SentryMan:
@rbygrave please review and deploy when you get time

PR #227, opened 2023-06-15
[HttpClient] Remove deprecated HttpClientContext, migrate to HttpClient

rob-bygrave:


SentryMan:
are we removing all the deprecated stuff now?

SentryMan:
I'd say V2, since this is a real breaking change

PR #229, opened 2023-06-19
Extendable Clients

SentryMan:
Now will generate the extended interface's http method implementation.

- adds a parameter to `@Client` that will disable generation is desired
- modify generation to read super client interface methods


SentryMan:
@rbygrave plz review and deploy when you get time

rbygrave (approved):

PR #230, opened 2023-06-24
(Javalin Jsonb) Add support for disableDirectWrites to NOT write directly to Jetty HttpOutput Stream

SentryMan:
Fixes wrong headers being added when body content exceeds a certain size. Also adds a flag to disable Jsonb direct writes.


rbygrave:
I would have been good to have a failing test case. Behaviour of Javalin can change and also if helps explain the change to future devs looking at why this PR came to be.

SentryMan:
In this case, it disconnects from Javalin even more, as we skip Javalin's CompressedOutputStream and lean more on Jetty directly. Which means it's less likely for Javalin behavior changes to affect serialization.

PR #233, opened 2023-07-08
(nima-generator) Support `InputStream` Return Types

SentryMan:
- use `transferTo` to write input stream contents to the nima output stream
- can use custom Media Types
- can handle checked exceptions thrown by controllers


SentryMan:
I'm good with this

PR #234, opened 2023-07-13
Add Accept Language header to validation

SentryMan:
In preparation for Avaje Validation:

- Modifies Validator interface to accept language headers and adds a method to resolve locales
- Modify Hibernate Validator to match
- Modify Generation to extract language header and groups for validation


SentryMan:
> ## Sonatype Lift is retiring

oof, we only just got it

SentryMan:
oh yeah it fails to build because the hibernate validator hasn't been deployed

rbygrave (approved):

PR #236, opened 2023-07-19
Nima - When returning JSON as String assume the string is raw json

rbygrave:
I think this is the right thing to do.  If so, should check this for Javalin.


SentryMan (approved):
oh, makes sense

PR #238, opened 2023-07-21
Deprecate the old reactive Helidon, effectively replaced by Nima going forward

rbygrave:


SentryMan:
we gonna rename the nima module?

PR #249, opened 2023-07-26
Change ValidationException errors to be a List<ValidationException.Error>

rbygrave:
This is effectively an API breaking change from Map -> List + explicit type

The ValidationException.Error type is fairly opinionated. If that doesn't match application requirements the error handler needs to obtain the cause and pull the constraint violation errors from that to build its error response payload message

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/ValidationException.java
+++ b/http-api/src/main/java/io/avaje/http/api/ValidationException.java
@@ -1,6 +1,7 @@
 package io.avaje.http.api;
 
-import java.util.Map;
+import java.util.ArrayList;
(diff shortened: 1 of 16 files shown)


SentryMan:
Error is pretty generic for a name

rbygrave:
> Error is pretty generic for a name

Yes, I'm keen to get a better name. Hmm.  `Violation` | `ViolationMessage` | `ErrorMessage` ...

SentryMan:
I vote violation

rbygrave:
> I vote violation

Actually I like Violation better than ViolationMessage too ... doing that ...

PR #254, opened 2023-07-30
Add `@ExceptionHandler`

SentryMan:
Adds a new exception handler annotation for generating error handler adapters.

- Adds new annotation
- Modify Javalin/Helidon generation to read the annotation and generate classes

relies on #252

### DIFF
--- a/tests/test-nima-jsonb/src/main/java/org/example/TestController.java
+++ b/tests/test-nima-jsonb/src/main/java/org/example/TestController.java
@@ -5,6 +5,8 @@ import java.util.List;
 import java.util.Set;
 
 import io.avaje.http.api.*;
+import io.helidon.nima.webserver.http.ServerRequest;
+import io.helidon.nima.webserver.http.ServerResponse;
 
 @Path("test")
 @Controller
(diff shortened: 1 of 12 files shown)


tests/test-nima-jsonb/src/main/java/org/example/TestController.java
```diff
@@ -53,14 +55,30 @@ String enumQueryImplied(String s, @QueryParam ServerType type) {
...
   InputStream stream(InputStream stream) throws Exception {
     return stream;
   }
 
-
   @Post("/strBody")
   String strBody(@BodyString String body) {
     return body;
   }
+
```
rob-bygrave, line 67: Lets merge this. I'm just thinking that in general exception handling is "global" and currently we attach it to controllers.

I'm thinking that ideally exception handling is added to a "root controller" or perhaps ... we allow target=class for `@ExceptionHandler` as a kind of "Exception Handler controller" that we expect all exception handlers to be added to (and not add them to controllers per se).

SentryMan: I could add a `@ControllerAdvice` annotation if you want

rob-bygrave: Hmm, let's not add another annotation just let.

rob-bygrave: > add a @ControllerAdvice annotation if you want

FYI:   Just to say we can just have a `@Controller` that _ONLY_ has `@ExceptionHandler` methods on it and no `@Path` so at this stage I think I am happy with that and I think that makes decent sense.

PR #255, opened 2023-07-30
Add Javalin `@Before`/`@After` and Helidon `@Filter`

SentryMan:
- Adds new annotations 
- Modifies Generators to use them

`@Filter` annotation translates to Helidon `Filter` and Javalin `before` handlers

relies on #254


http-generator-helidon/src/main/java/io/avaje/http/helidon/Filter.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method as a Helidon Filter
+ */
+@Target(METHOD)
+@Retention(SOURCE)
+public @interface Filter {}
```
rob-bygrave, line 14: I'd expect the Before/After/Filter annotations to be in avaje-http-api.

Why is this not in avaje-http-api?

SentryMan: It basically makes it clear that the `Before/After` functionality is Javalin only, and vice versa for Helidon Filter.

rob-bygrave: But putting it into the generator is worse no?  We don't want the application controllers etc to have a dependency on the generators.

We might have a problem with Before/After being Javalin specific (at the moment) and Filter being Helidon specific but maybe we have options for dealing with that - like an explicit exception when defined but not supported etc.

These annotations can't be in the generators though right?

rob-bygrave: For myself, I'd rather these annotations where in avaje-http-api ... and we dealt with the issue of being Javalin or Helidon specific in some other way (like the generators saying that we can't use those annotations with that particular target).

SentryMan: We already say that the generators should be added as provided dependencies, with the annotations designated with `Retention.SOURCE` I don't see an issue, since there is still no hard dependency on the generators

SentryMan: I suppose I'm not keen on having "useless" classes hanging around and potentially causing confusion.

SentryMan: there is also the matter of javalin adding more lifecycle methods `beforeMatched`/`afterMatched` so the amount of "useless" annotations will increase.

rob-bygrave: How about we move `@Filter` into avaje-http-api.  That will keep me happy wrt Helidon Nima and I think `@Filter` is _"generally correct / expected"_.

We can keep the Javalin annotations where they are for now, maybe move them into a separate avaje-javalin-api later.

http-generator-helidon/src/main/java/module-info.java
```diff
@@ -1,10 +1,13 @@
 module io.avaje.http.nima.generator {
 
-  provides javax.annotation.processing.Processor with io.avaje.http.generator.helidon.nima.NimaProcessor;
+  exports io.avaje.http.helidon;
```
rob-bygrave, line 3: And we remove this with `@Filter` in the API

SentryMan:
You drive a hard bargain

PR #263, opened 2023-08-07
Add HttpStatus enum for status code

rbygrave:
Not ideal in the sense that it has the `__not_set(0)` entry.  The default status codes depend on the HTTP method + if the return body is empty (void methods).

An alternative is to have a separate annotation like `@Status` or `@ResponseStatus` and remove the statusCode attributes from `@Produces` and `@ExceptionHandler` (and then the __not_set entry could be remove from the HttpStatus enum).

Also not ideal in that this PR just copies the enum for the generator.


http-api/src/main/java/io/avaje/http/api/HttpStatus.java
```diff
@@ -0,0 +1,84 @@
+package io.avaje.http.api;
+
+public enum HttpStatus {
```
SentryMan, line 3: idk, most microframeworks provide their own HTTP status enum

rbygrave: But how would they use it on an avaje annotation?

Currently we use an int meaning folks need to know the http status codes which is maybe OK but in theory this helps them.  Is there a better way?

SentryMan: Idk, I like using int

rbygrave: OK, we can stick with int.

rob-bygrave: I have the thought of adding a java comment in the generated code [where the generated code sets the status] that translates the status code int into the reason description (for the purpose of helping the person reading the generated code who may not know the status code being used).

That way we use int, but at least the generated code might be more readable for devs reading the [generated] code later.

SentryMan: sounds good to me

PR #264, opened 2023-08-08
[http-client] Generate VarArgs Correctly

SentryMan:
Adds a way to detect if an element is a varArg, and generates the implementation with a vararg param instead of an array.


SentryMan:
I've never had to use regex this much until I started working on these libs. It has turned out to be much more useful for AP than I had imagined. In any case, I'm good with this.

rob-bygrave (approved):

PR #266, opened 2023-08-08
Only generated response status code when statusCode > 0

rob-bygrave:
Support use case that requires programmatic control of status code but non-void controller method. This might be marginal but I don't think its a negative.


SentryMan (approved):
expecting a negative status code one of these days?

rob-bygrave:
> expecting a negative status code one of these days?

Not really. The use case is pretty marginal in that if folks need programmatic control over the statusCode then they can take complete control, use a method that takes the ServerResponse / Context, returns void ... and completely control how the response is written.

This prevents any extra `response.setStatus(...)` being generated if for some reason they want that control but actually still return a non-void response.

Its very marginal yeah.

PR #270, opened 2023-08-12
Make `ValidationException` Immutable/Support Multi-Value Form Parameters.

SentryMan:
There's no reason I can see to have the exception be mutable.

Also form parameters can be multi-valued

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/ValidationException.java
+++ b/http-api/src/main/java/io/avaje/http/api/ValidationException.java
@@ -49,12 +49,12 @@ public class ValidationException extends IllegalArgumentException {
   }
 
   /** Return the suggested HTTP status to use in the response. */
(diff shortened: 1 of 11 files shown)


http-api/src/main/java/io/avaje/http/api/ValidationException.java
```diff
@@ -101,20 +92,5 @@ public String getField() {
     public String getMessage() {
       return message;
     }
-
```
rob-bygrave, line 104: I originally left this mutable with getters/setters such that it would work out of the box with Jackson.

I'm thinking we can merge this in though and see how it goes.

SentryMan: you only really need getters in that case

PR #271, opened 2023-08-13
Change Violation to use record style accessors

rbygrave:
This makes it consistent with record type which it could be when this library becomes Java 17+


SentryMan:
see I tried that, but that doesn't work with Jackson.

PR #274, opened 2023-08-24
[http-client] Exception Mapper/Async Retry

### LINKED ISSUE #273: [http client] Add support for "error decoder" Function<HttpException, DistanceException>
Currently we either 

1: allow `io.avaje.http.client.HttpException` to be thrown, catch that and decode if necessary like:

```java
    } catch (HttpException e) {
      // handle error, convert to the API specific ErrorResponse
      ErrorResponse err = e.bean(ErrorResponse.class);
      UUID id = err.getId();
      Integer code = err.getCode();
      String errorMessage = err.getError();
    }
```

2: register a `RequestIntercept` and in `afterResponse()` decode the response and throw an Exception as desired

This proposal is to make this easier / nicer by:

**A)** Allowing us to register a "error decoder" that implements `Function<HttpException, Throwable>` that converts the HttpException into an application specific exception.  This would be registered with the HttpClient.Builder.

**B)** For a `@Client` annotated interface, when a method in the interface is declared to throw an Exception, generate code that converts HttpException to that interface declared Exception [look at the constructors on the exception and generate code that converts HttpException into that]. 

e.g.
```java
@Client
public interface Simple {

  @Get("users/{user}/repos")
  List<Repo> listRepos(String user, String other) throws MyApplicationException;
```
... then the generated client generates code that catches HttpException and converts it into MyApplicationException.

SentryMan:
Fixes that long-standing async retry capability gap and adds exception mapping for requests

- now can configure a global exception mapper that maps `HttpException` to other kinds of exception
- can set error handlers on a per-request basis to override the global mapper
- Modify generator to read `throws` statement on Client interfaces, if the Exception has a constructor that accepts  `HttpException`, an error mapper will auto generate for the request
- Async Requests now propagate connection errors as `HttpException` matching sync behavior
- Async requests can now use retry handlers

```
@Client
interface ApiClient {

  @Get("/mapped")
  String mapped(@Header("Accept") String accept) throws MappedException;
}
class MappedException extends RuntimeException {

  public MappedException(HttpException e) {}
}
```
The following impl method will generate

```
  // GET /mapped
  @Override
  public String mapped(String accept) {
    return client.request()
      .header("Accept", accept)
      .path("mapped")
      .errorMapper(MappedException::new)
      .GET()
      .asPlainString().body();
  }
```

fixes #273


http-client/src/main/java/io/avaje/http/client/DHttpClientRequest.java
```diff
@@ -545,13 +555,20 @@ public <T> HttpResponse<T> handler(HttpResponse.BodyHandler<T> responseHandler)
...
   private <T> HttpResponse<T> sendWith(HttpResponse.BodyHandler<T> responseHandler) {
     context.beforeRequest(this);
     addHeaders();
-    final HttpResponse<T> response = performSend(responseHandler);
+    final HttpResponse<T> response;
+    if (errorMapper != null) {
+      try {
+        response = performSend(responseHandler);
+      } catch (final HttpException e) {
+        throw errorMapper.apply(e);
```
rbygrave, line 567: This `errorMapper.apply()` isn't where I was expecting it to be. 

I was thinking it would be part of `checkMaybeThrow()` ... so I'm pondering this. This does not look right in that we want to let the response be decoded if necessary and HttpException is thrown later?

I'll pull this and see how it works and why it's not what I was expecting - looks cool otherwise.

SentryMan: This is so it'll work for connection exceptions

SentryMan: Hmm, wait you might be right

rbygrave: Yeah, we need the mapping to occur also in the checkResponse() and checkMaybeThrow().

See the additional commits at: https://github.com/avaje/avaje-http/pull/275

PR #277, opened 2023-08-24
[http client] HttpException add isPlainText() and contentType()

rbygrave:
Helper methods to make it easier for use cases that need logic in reading the HttpException that could be text or json etc


SentryMan (approved):
LGTM

PR #281, opened 2023-08-26
[http-client] Validate Module

SentryMan:
Fail compilation on modular projects if people forget to add the required `provides` clause.

### DIFF
--- a/http-generator-client/src/main/java/io/avaje/http/generator/client/ClientProcessor.java
+++ b/http-generator-client/src/main/java/io/avaje/http/generator/client/ClientProcessor.java
@@ -49,7 +49,9 @@ public class ClientProcessor extends AbstractProcessor {
 
   @Override
   public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment round) {
-    final var platform = platform();
+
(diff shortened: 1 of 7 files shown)


http-generator-client/src/main/java/io/avaje/http/generator/client/ClientProcessor.java
```diff
@@ -6,6 +6,7 @@
 import static io.avaje.http.generator.core.ProcessingContext.typeElement;
 
 import java.io.IOException;
+import java.security.Identity;
```
rbygrave, line 9: I'll remove this import.

http-generator-client/src/main/java/io/avaje/http/generator/client/ClientProcessor.java
```diff
@@ -49,7 +50,9 @@ public synchronized void init(ProcessingEnvironment processingEnv) {
 
   @Override
   public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment round) {
-    final var platform = platform();
```
rbygrave, line 52: I'll check indent

PR #285, opened 2023-09-12
Add BigInteger param  type mapping

SentryMan:
now can convert query params into BigIntegers


http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
```diff
@@ -131,6 +132,18 @@ public static boolean asBool(String value) {
     return asBoolean(value);
   }
 
+  /**
+   * Convert to BigInteger (not nullable).
+   */
+  public static BigInteger asBigInteger(String value) {
```
rob-bygrave, line 138: There are 2 methods for each type in this PathTypeConversion.  the asType and toType methods ... with the toType methods allowing null.

So the `toBigInteger()` method looks to be missing.  It would be used by nullable query parameters (and the asBigInteger() would be used for the non-nullable path parameters).

rob-bygrave (approved):

PR #302, opened 2023-10-14
Enable annotation processors in JDK 22

SentryMan:
They disabled annotation processors by default in JDK 22

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,11 +59,22 @@
     <profile>
       <id>jdk19plus</id>
       <activation>
-        <jdk>[20,21]</jdk>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -56,15 +56,6 @@
         <module>tests</module>
       </modules>
     </profile>
-    <profile>
```
rob-bygrave, line 59: This is not correct. We have an issue with the current build that it does not build and test `http-generator-helidon`. I'm fixing that.  After that we should find the EA JDK 22 build failing ... and then should be able to fix that with a bump to the parent pom.

PR #323, opened 2023-10-18
[http-client] Allow paths to be loaded from System Properties/Avaje Config

SentryMan:
Now can dynamically determine the paths at runtime using system properties or avaje config.

With this, we can perform tomfoolery such as:

```java
@Client
public interface TitanFall {

  @Get("/${titan}/${drop.point}")
  Titan titanfall();
}
```

and the following will generate 

```java
@Generated("avaje-http-client-generator")
public class TitanFallHttpClient implements TitanFall {
  private final HttpClient client;
  public TitanFallHttpClient(HttpClient client) {
    this.client = client;
  }
                                //if available will use Config.get instead 
  private static final String DROP_POINT = System.getProperty("drop.point");
  private static final String TITAN = System.getProperty("titan");
  // GET /${titan}/${drop.point}
  @Override
  public Titan titanfall() {
    return client.request()
      .path(TITAN).path(DROP_POINT)
      .GET()
      .bean(Titan.class);
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Use source code generation to adapt annotated REST controllers `@Path, @Get, @Po
 
 - Lightweight (65Kb library + generated source code)
 - Full use of Javalin or Helidon SE/Nima as desired
-- Bean Validation of request bodies supported
+- Bean Validation of request bodies supported (validation groups supported as well)
 
 ## Add dependencies
 ```xml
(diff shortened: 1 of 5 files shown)


rbygrave (approved):

rbygrave:
BTW: I think this is a pretty cool and useful feature. I'd not be surprised to see quite a few folks use it in time.

PR #324, opened 2023-10-18
[http-client] Reset base if an http scheme is detected

SentryMan:
now if we do something like:

```
  @Override
  public Titan titanfall() {
    return client.request()
      .path("https://something.com").path(DROP_POINT)
      .GET()
      .bean(Titan.class);
  }
```
the base URL  will be switched to `https://something.com`.

- `path`  will reset the URL buffer if an HTTP scheme is detected in the given string

### DIFF
--- a/http-client/src/main/java/io/avaje/http/client/DUrlBuilder.java
+++ b/http-client/src/main/java/io/avaje/http/client/DUrlBuilder.java
@@ -22,6 +22,9 @@ final class DUrlBuilder implements UrlBuilder {
 
   @Override
   public UrlBuilder path(String path) {
+    if (path.startsWith("http") && path.contains("://")) {
+      return url(path);
+    }
     buffer.append("/").append(path);
     return this;
   }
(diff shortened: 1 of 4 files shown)


http-client/src/main/java/io/avaje/http/client/DUrlBuilder.java
```diff
@@ -15,13 +15,21 @@ final class DUrlBuilder implements UrlBuilder {
...
+
     buffer.append(url);
     return this;
   }
 
   @Override
   public UrlBuilder path(String path) {
+    if (path.startsWith("http") && path.contains("://")) {
+      buffer.setLength(0);
+      return this;
```
rbygrave, line 31: this does not look right?  the path is not appended before the `return this` ?

SentryMan: oh woops

rbygrave (approved):

PR #332, opened 2023-10-21
[http-generator-core] OpenAPI Schema nullable property - Added support for NotEmpty and NotBlank 

mrguamos:
Currently, only `NotNull` is supported


SentryMan (approved):
LGTM

PR #333, opened 2023-10-22
[http-generator-core] added required props

mrguamos:
This is to add support for `required` label on fields that are not nullable (any of "NotNull", "NotEmpty", "NotBlank") 

```json
  "components": {
    "schemas": {
      "Provider": {
        "required": [
          "firstName",
          "lastName"
        ],
        "type": "object",
        "properties": {
          "firstName": {
            "type": "string",
            "nullable": false
          },
          "middleName": {
            "type": "string"
          },
          "lastName": {
            "type": "string",
            "nullable": false
          }
        }
      }
    }
  }
```

Sample redoc

image


SentryMan (approved):
LGTM

SentryMan:
your branch is out of date, update it and we should be good

PR #342, opened 2023-11-01
Javadoc for AvajeJavalinPlugin and tidy javalin ControllerWriter only

rob-bygrave:


SentryMan (approved):

rob-bygrave:
FYI: I've added a comment to the Javalin PR that changed Plugin into an abstract type, will see what the reply there is.

PR #346, opened 2023-11-14
Test Javalin 6 `beta.2`

SentryMan:
By the way, can we get an RC candidate for those lads who want to use this beta version with avaje?


rob-bygrave:
> lads

lads **_AND_** ladies :)

rob-bygrave (approved):

SentryMan:
lads and ladettes

PR #356, opened 2023-12-12
Add `toEnum`

### LINKED ISSUE #354: PathTypeConversion is missing toEnum() method for nullable enum as query parameter

SentryMan:
fixes #354


SentryMan:
let it be known I was first

PR #393, opened 2024-02-23
Javadoc badges

SentryMan:


http-client/src/main/java/io/avaje/http/client/HttpClient.java
```diff
@@ -69,7 +70,6 @@ static Builder builder() {
   HttpClientRequest request();
 
   /**
-   * Return a UrlBuilder to use to build an URL taking into account the base URL.
```
rbygrave, line 72: Removing this javadoc?

SentryMan: oh woops

rob-bygrave (approved):

PR #395, opened 2024-02-24
[http-client] Support JDK http requests

SentryMan:
Now we can use the built in jdk request classes to construct requests


SentryMan:
ehhh, now that I think about it, this isn't really needed

PR #414, opened 2024-04-09
Support Byte-Buddy / Mockito with JDK 23 via -Dnet.bytebuddy.experimental=true

rob-bygrave:


SentryMan (approved):

rob-bygrave:
Its very tempting to include this "fix" / maven-surefire-plugin configuration in the parent pom. So more like:

- Update io.avaje:junit to include the latest for mockito and byte-buddy
- Update the parent pom to have this configuration for maven-surefire-plugin and maven-failsafe-plugin enabled by default

SentryMan:
> * Update io.avaje:junit to include the latest for mockito and byte-buddy

I'd add dependabot to it but I don't know where avaje-junit is

SentryMan:
*checks notes* It seems that avaje-junite merely aggregates some common testing libs. Do you think we would get better mileage by directly declaring and letting dependabot take care of things?

PR #427, opened 2024-05-02
Fix 204 Deserialization error.

### LINKED ISSUE #426: [HTTP-Client] 204 deserialization error
[Essentially Copied from discord]

Suppose an api call to a server results in a 204. `.GET().as(MyBean.class)` tries to deserialise the body (of which is empty), resulting in an error during deserialization 

What I would've expected is:
1. Just don't deserialise when it's a 204 or you know the body is empty in the first place.
2. Only deserialise when `.body()` is called. That way it's down to the implementation to know when it should and shouldn't expect a body. So for example, in a 204, I shouldn't call `.body()` but in other scenarios I should probably expect the body to be present.

SentryMan:
Should fix #426


rbygrave:
Replaced by https://github.com/avaje/avaje-http/pull/430

PR #436, opened 2024-05-18
[http-client] Fix BeanParam on Eclipse

SentryMan:
I guess eclipse prints out the entire method signature when doing `ExecutableElement#toString()`

```java
  @Override
  public HttpResponse<InputStream> call1(Params s) {
    return client.request()
      .path("hmm")
      .queryParam("param", s.public java.lang.String getParam() )
      .queryParam("named", s.public java.lang.String getParam2() )
      .header("Head", s.public java.lang.String getHead() )
      .queryParam("publicQuery", s.publicQuery)
      .GET()
      .asInputStream();
  }

```


http-generator-client/src/test/java/io/avaje/http/generator/client/clients/Params.java
```diff
@@ -0,0 +1,49 @@
+package io.avaje.http.generator.client.clients;
+
+import io.avaje.http.api.Cookie;
+import io.avaje.http.api.Header;
+import io.avaje.http.api.QueryParam;
+
+public class Params {
```
rbygrave, line 7: This does not add anything to the test coverage?

We have `@BeanParams` in tests/test-client-generation in JunkApi and CommonParams and if we needed more coverage we would add it there?

I'll remove this I think.

SentryMan: Perhaps so, but these generator tests are what allow me to properly debug the processor with my IDE.

PR #442, opened 2024-06-02
[Http-Client] Use virtual thread executor by default

SentryMan:
When on JDK 21+, the builder will default to the `VirtualThreadPerTask` Executor for the underlying `HttpClient`.

### DIFF
--- a/http-client/src/main/java/io/avaje/http/client/DHttpClientBuilder.java
+++ b/http-client/src/main/java/io/avaje/http/client/DHttpClientBuilder.java
@@ -105,7 +105,7 @@ final class DHttpClientBuilder implements HttpClient.Builder, HttpClient.Builder
     }
     if (executor != null) {
       builder.executor(executor);
-    } else {
+    } else if (Integer.getInteger("java.specification.version") >= 21) {
       try {
         ExecutorService virtualExecutorService =
             (ExecutorService)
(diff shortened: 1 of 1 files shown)


rob-bygrave:
Did someone request this change or do we just think it's a good idea?

SentryMan:
I just think it's cool

SentryMan:
@rbygrave yea or nay?

rbygrave:
> yea or nay?

As I see it, this looks like a good idea but I'm not quite 100% sure myself. So I'm sitting on the fence waiting for someone to hit a use case where they want this as the default. I guess the JDK HttpClient has to be more conservative which is perhaps why this isn't the default in the JDK. 

If we merge this so that it becomes the default behaviour then it is hard to "undo" that in that doing so would be a "silent regression" type of change. 

In this sense we need to be 100% sure when/if we merge this.  For myself I'm not quite 100% yet.  Can we keep this on hold for now?

SentryMan:
I'm hearing nay (at least for now)

rbygrave:
Ok, but if we leave the PR open ... then other people can see it and add a comment. If we delete the PR it's now really unlikely to happen?

re-thc:
Maybe best to wait until [JEP 491](https://openjdk.org/jeps/491) gets merged

SentryMan:
> Maybe best to wait until [JEP 491](https://openjdk.org/jeps/491) gets merged

I mean the current implementation of HTTP client doesn't use synchronized anyways

rbygrave:
Ok, I'm thinking maybe we should merge this in with the thinking that this should be a reasonable default in a Java 21+ world. 

That is, bump the next released version to `3.0` and merge this in.

rbygrave (approved):

PR #498, opened 2024-09-05
Add Sigma Lambda Generator

### LINKED ISSUE #496: Add an AWS Lambda Generator
People tend to make apis with lambdas, why not have the plumbing be automated by a code generator?

People should be able to write:
```java
@Controller("/widgets")
public class WidgetController {
  private final HelloComponent hello;
  public WidgetController(HelloComponent hello) {
    this.hello = hello;
  }

  @Get("/{id}")
  Widget getById(int id) {
    return new Widget(id, "you got it"+ hello.hello());
  }

  record Widget(int id, String name){};
}
```

And have the lambda (de)serialization and routing code generated.

We're gonna need:

- [x] Some sort of Router Class that'll match the routes
- [x] A Json Interface for (de)serializing the body
- [x] Interfaces for wiring
- [x] support for both API Gateway and ALB

After all of that, we can start generating the code to hook into this.

SentryMan:
Creates a generator for avaje-sigma. 

Resolves #496


SentryMan:
@rbygrave please review and deploy when you get time

rbygrave (approved):

PR #506, opened 2024-10-09
Add roles support for Helidon

re-thc:
Helidon has security providers e.g. OIDC. These integrate with WebServer Security as `SecurityHandler`s with the shorthand `SecurityFeature.rolesAllowed(<roles>)` and the authorization provider will then validate the role before delegating to the actual handler.

This PR takes Avaje's `@Roles` and populates them and the rest is up to Helidon.


SentryMan (approved):

rbygrave (approved):

SentryMan:
@rbygrave by the way, can you release an RC?

PR #509, opened 2024-10-28
Reproducible build

SentryMan:
configures build to be [reproducible](https://maven.apache.org/guides/mini/guide-reproducible-builds.html)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,5 +33,5 @@ jobs:
         env:
           JAVA_VERSION: ${{ matrix.java_version }}
         run: |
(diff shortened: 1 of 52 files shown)


SentryMan:
seems it works on both mac and windows, but not in the build

SentryMan:
I was reading [about all the benefits](https://reproducible-builds.org/) of reproducible builds, and I couldn't find any reason not to try seeing if the libraries here can have reproducible builds. For most of them it was a simple pom configuration.

SentryMan:
I suppose a more immediate tangible benefit is seeing and understanding where the randomness in the code-gen comes from.

SentryMan:
You have to admit it's curious that that generator core is the one causing the problem. This raises some very interesting questions.

- Is it an issue with the prisms generation? 
- If so, why does it fail here and not on the other repos? 
- Is http doing something else with the build that causes the core jar in particular to have different results every time?
- Is it the weird shading thing we have going on?

SentryMan:
well I'm stumped

PR #515, opened 2024-11-25
[Jex] Modify generated code for Jex controllers to use handler methods

rbygrave:
This will make it easier to support an HTMX extension for jex


SentryMan (approved):
LGTM

PR #517, opened 2024-11-26
Update Jex Generation

SentryMan:
- Generates code for the new version of Jex
- adds OPTIONS HTTP annotation


SentryMan:
@rbygrave should be good to go now

rbygrave:
Yeah. I think the only issue is that the Java 17 build is broken due to it depending on jex generator that is Java 21+ only now right?

SentryMan:
yeah that's the reason

rbygrave:
So remove that dependency or make test-client-generation a Java 21+ only test ?

rbygrave (approved):

PR #522, opened 2024-12-02
[Jex-Generator] Jsonb Type generation

SentryMan:
Adds the explicit Jsonb type generator we have for helidon/javalin to the jex generator


SentryMan:
@rbygrave it seems the failing test is unrelated to my changes

rbygrave:
So with the test in question we have an interesting overlap for routes:
```
    routing.get("/common/name/{name}", this::_name);
    routing.get("/common/{id}/{name}", this::_p2);
```
So perhaps with EA builds the ordering here is changed (compile order of methods is not preserved?) or ... perhaps the routes lose ordering in the internal RoutesBuilder which uses an EnumMap.

I'll change that EnumMap to a LinkedHashMap and see ...

PR #523, opened 2024-12-02
[jex generation] fix newline in generated code

rbygrave:


http-generator-jex/src/main/java/io/avaje/http/generator/jex/ControllerMethodWriter.java
```diff
@@ -256,7 +256,6 @@ private void write(boolean requestScoped) {
       } else {
         writer.append(indent);
         writeContextReturn(responseMode, "result");
-        writer.eol();
```
rbygrave, line 259: This is the change - move this to be inside the writeContextReturn()

PR #536, opened 2024-12-17
Add AWS CognitoAuthTokenProvider module

rbygrave:
An HttpClient AuthTokenProvider for AWS Cognito


rbygrave:
It was missing the `<packaging>pom</packaging>`

SentryMan (approved):

rbygrave:
Oh wow, I didn't know we could do `<module>aws-cognito/http-client-authtoken</module>` ... now I'm thinking we could/should release this independently by itself on its own cadence.  So my thinking here is to move this under `test21` so that it isn't released with everything else via `-Pcentral` central profile.

PR #550, opened 2025-01-18
added support for Optional Request Parameters

LukasDetermann:
I wanted to contribute and asked @SentryMan for a good starter feature. I tried to emulate the style.

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
@@ -4,6 +4,7 @@ import java.math.BigDecimal;
 import java.time.*;
 import java.math.BigInteger;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


SentryMan (approved):
LGTM

rbygrave:
Thanks for the PR, nice work. Note that I have submitted a PR to fix the issues/bugs that are in this PR/change.

The main thing to note is that in the past I've allowed some PRs to be merged when they didn't originally actually have tests to assert the behaviour and so I've inadvertently allowed the standards to drop in that regard. Instead I've added those tests to the PR's myself or done a followup PR that adds the tests - we should not do this going forward.

That is, any new feature or bug fix or change should include tests that actually exercise that feature. It is not sufficient for something to compile, we actually need tests to assert against the expected behaviour. In this case those tests were missing from this PR. In adding those tests we would have hit the issues with this change (around handling null, and that the tests had invalid URLs.

Cheers, Rob.

PR #555, opened 2025-01-26
Fix PR #550 Optional type wrapping of query parameters

rbygrave:
- Adds tests that were missing in HelloControllerTest that actually exercises the feature
- Fixes the controller URLs removing incorrect {myOptional} from urls
- Fixes PathTypeConversion to handle null values
- Adds missing javadoc on new public methods of PathTypeConversion
- Adds some useful unit tests for TypeMap.OptionalHandler


SentryMan (approved):
long time no see

PR #557, opened 2025-01-27
[client] Breaking change - Rename @Client.Import(types=...) to @Client.Import(value=...).  That is, rename types -> value

SentryMan:
Now can do `@Client.Import(CustomType.class)`

--------

This is a breaking change

From:
```java
@Client.Import(types = OtherApi.class)
```

To:
```java
@Client.Import(value = OtherApi.class)
```

... and because it is value it can be now written as:

```java
@Client.Import(OtherApi.class)
```

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/Client.java
+++ b/http-api/src/main/java/io/avaje/http/api/Client.java
@@ -53,6 +53,6 @@ public @interface Client {
     /**
      * Client interface types that we want to generate HTTP clients for.
      */
-    Class<?>[] types();
+    Class<?>[] value();
   }
 }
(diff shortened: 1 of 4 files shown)


SentryMan:
also shouldn't we be doing `3.0-RC`s?

rbygrave (approved):

rbygrave:
> also shouldn't we be doing 3.0-RCs?

Yes, these 2.9-RCs are really 3.0-RCs.  We started on 2.9.x and I didn't change it ... but maybe we should.

SentryMan:
time for 3.0-RC1

PR #563, opened 2025-02-10
Update jex context package

SentryMan:
If https://github.com/avaje/avaje-jex/pull/182 get's merged, this is required to properly generate jex handlers


rbygrave (approved):

SentryMan:
I think we're good for 3.0

PR #569, opened 2025-02-24
[generators] Remove jsonb generation if no jsontypes available

SentryMan:
Now will not generate Jsonb imports/constructor if no json types are detected on the controller


rbygrave:
I've refactored, pulling the logic into a new JsonBDetect. If you are happy then we can merge this I reckon.

SentryMan:
I'm good with it

rbygrave (approved):

PR #574, opened 2025-03-10
Add disableJsonB option to NOT use avaje-jsonb in code generation eve…

rbygrave:
…n if it's in the classpath


SentryMan (approved):
LGTM, but what brought this on?

rbygrave:
Had someone creating a Nima rest api, it was using JAX-B generated classes for request/response types ... and those used java.util.Calendar [and also no setters for the collections].  So there wasn't a relatively quick/easy way to turn off the use of avaje-jsonb in the generated code and just use Jackson [apart from removing avaje-jsonb entirely from the classpath].

So this is there to support that type of case, where someone hits some issue and just wants to "downgrade" to using Jackson [perhaps temporarily until any issue with avaje-jsonb is sorted].

PR #578, opened 2025-03-17
[jex-generator] use body stream methods

SentryMan:
uses the body methods of jex. (Direct reading jsonb has the disadvantage of throwing an exception when the stream has been read)


rbygrave (approved):

SentryMan:
we might be good for 3.1

PR #581, opened 2025-03-21
Add compiler error for GET json bodies

### LINKED ISSUE #580: @Get with a bean generates code to deserialize as json from request body (but there is no request body)
Given a `GET` like this:
```java
    @Get("/search/criteria")
    public SearchResult findByCriteria(SearchCriteria criteria) {
```
... the generated code looks to read critera as json content from the request body - but its a `GET` and so **_there is no request body_** so we get a strange json deserialization error. 

### Suggested Fix:
The proposed fix, is that the generated code should see this is a `GET` and so it must look to populate the criteria bean using path parameters instead.

Note that currently for `POST` and `@Form`, criteria would be populated from _form parameters_ so this is pretty similar to what should be generated [but use path parameters rather than form parameters]. Example:
```java
    @Post("/search/criteria")
    public SearchResult findByCriteria(@Form SearchCriteria criteria) {
```

SentryMan:
Resolves #580

### DIFF
--- a/http-generator-core/src/main/java/io/avaje/http/generator/core/MethodReader.java
+++ b/http-generator-core/src/main/java/io/avaje/http/generator/core/MethodReader.java
@@ -60,7 +60,7 @@ public class MethodReader {
     this.bean = bean;
     this.element = element;
     this.actualExecutable = actualExecutable;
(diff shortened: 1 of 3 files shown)


rbygrave (approved):

rbygrave:
Awesome, thanks

PR #585, opened 2025-03-26
Generate Test Clients during test compilation 

### LINKED ISSUE #584: New annotation to generate a test http client interface for a Controller
Say we have a controller like below, and we want to test that controller:

```java
@Controller
@Path("/ping")
class PingController {

    @Produces("text/plain")
    @Get
    String ping() {
          return Instant.now().toString();
    }
}
```

In a test, we specify that we desire a http client interface generated for this controller like say:

```java
@GenerateHttpClient(controller=PingController.class, clientPattern="{}ClientApi")
```
... and that that generates a http client interface in test generated source like:

```java
@Client
@Path("/ping")
interface PingClientApi {

    @Get
    HttpResponse<String> ping();
}
```
Note that the response types on the client interface should probably use `HttpResponse` so that the tests can access the response statusCode and response headers. So return `HttpResponse<String>` rather than `String` etc.

We can now inject the PingClientApi in a test like:

```java
@InjectTest 
class MyPingControllerTest {

  @Inject PingClientApi pingClientApi;

    @Test
  void ping() {
        HttpResponse<String> res = pingClientApi.ping();

        assertThat(res.statusCode()).isEqualTo(200);
        assertThat(res.body()).isNotEmpty();
  }

}

```

What do we think about this idea?

SentryMan:
mostly resolves #584 with a few caveats

- won't generate client methods for controllers with a Context parameter
- won't generate for controllers that use API

There are some bugs to work out using this approach. It seems that during test compilation it causes the classes to get recompiled so inject complains with errors like  `No dependency provided for org.example.web.myapp.WebController on org.example.web.myapp.WebController$Route`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,6 @@ dependency-reduced-pom.xml
 *avaje-processors.txt
 *controllers.txt
 tests/test-sigma/io.avaje.jsonb.spi.JsonbExtension
+tests/test-sigma/*.txt
 tests/test-javalin-jsonb/*.txt
+tests/test-nima-jsonb/*.txt
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,7 @@ build/
 *.Module
 dependency-reduced-pom.xml
 .DS_Store
-tests/test-sigma/avaje-processors.txt
+*avaje-processors.txt
+*controllers.txt
 tests/test-sigma/io.avaje.jsonb.spi.JsonbExtension
+tests/test-javalin-jsonb/*.txt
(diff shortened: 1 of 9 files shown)


SentryMan:
It turns out that even trying the traditional way with an annotation doesn't work either. Might have to get creative

SentryMan:
Figured it out by writing the source into a text file in target during compilation then loading and writing directly during test compilation.

SentryMan:
let me actually write a test first

rbygrave (approved):

rbygrave:
> writing the source into a text file in target during compilation then loading ...

That was cunning!!  Well done!!

rbygrave (approved):

PR #596, opened 2025-04-24
Fix Helidon path openapi generation

### LINKED ISSUE #595: Path parameters are being output to the openapi.json as query params
We've been toying with using Helidon SE for developing some new services, however we take a fairly strict dependency on OpenAPI specs (for better or worse). After the compile phase I copy the generated `openapi.json` out to a project directory to be "vendored" and referenced for things like client generation and whatnot. 

What I found is that the openapi generator actually throws validation errors when using the Avaje HTTP generated openapi.json like the following:

```
> There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
   | Error count: 6, Warning count: 0
  Errors:
        -paths.'/api/v2/{depId}/definitions/achievement'. Declared path parameter depId needs to be defined as a path parameter in path or operation level
...
```

Looking a little closer, I found that it was indeed outputting the path parameters as query params. Like the following example:

```java
@Get("/find/{type}")
List<Contact> findByType(String type, @QueryParam String lastName) {
    // ...
}
```

The expected openapi spec should specify `type` as a `path` parameter, and `lastName` as a `query` parameter.

But it outputs:

```json
"/contacts/find/{type}" : {
	"get" : {
		"tags" : [
			
		],
		"summary" : "",
		"description" : "",
		"parameters" : [
			{
				"name" : "type",
				"in" : "query", <------- should be path
				"schema" : {
					"type" : "string"
				}
			},
			{
				"name" : "lastName",
				"in" : "query",
				"schema" : {
					"type" : "string"
				}
			}
		],
```

SentryMan:
Resolves #595

### DIFF
--- a/http-generator-helidon/src/main/java/io/avaje/http/generator/helidon/nima/ControllerWriter.java
+++ b/http-generator-helidon/src/main/java/io/avaje/http/generator/helidon/nima/ControllerWriter.java
@@ -99,6 +99,9 @@ class ControllerWriter extends BaseControllerWriter {
     writeRoutes(methods);
     for (final ControllerMethodWriter methodWriter : methods) {
       methodWriter.writeHandler(isRequestScoped());
+      if (!reader.isDocHidden()) {
+        methodWriter.buildApiDocumentation();
+      }
     }
   }
 
@@ -108,9 +111,6 @@ class ControllerWriter extends BaseControllerWriter {
 
     for (final ControllerMethodWriter methodWriter : methods) {
       methodWriter.writeRule();
-      if (!reader.isDocHidden()) {
-        methodWriter.buildApiDocumentation();
-      }
     }
     writer.append("  }").eol().eol();
   }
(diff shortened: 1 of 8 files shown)


http-generator-helidon/src/main/java/io/avaje/http/generator/helidon/nima/ControllerWriter.java
```diff
@@ -108,9 +111,6 @@ private void writeRoutes(List<ControllerMethodWriter> methods) {
 
     for (final ControllerMethodWriter methodWriter : methods) {
       methodWriter.writeRule();
-      if (!reader.isDocHidden()) {
```
SentryMan, line 111: the problem was that it was setting the openApi information before the parameter  was determined to be a path

rbygrave (approved):

PR #600, opened 2025-04-29
Fix generated openapi.json parameter names

rbygrave:
```java

@QueryParam String lastName               // -> "lastName" (no bug here, this was OK)
@QueryParam("q-2") String param2          // -> "q-2"
@Header String contentLength              // -> "Content-Length" (because headers automatic names are Init-Camel)
@Header("x-oh") String otherHeader        // -> "x-on"

```


rbygrave:
Ok, I'll merge this in before a PR review because I'm wanting another RC release to test out. If you are not liking this change @SentryMan  we can adjust again.

PR #606, opened 2025-05-21
[client] support custom classloader

SentryMan:
now can load client impls in all sort of weird situations

### DIFF
--- a/http-client/src/main/java/io/avaje/http/client/DHttpApi.java
+++ b/http-client/src/main/java/io/avaje/http/client/DHttpApi.java
@@ -1,29 +1,27 @@
 package io.avaje.http.client;
 
-import io.avaje.applog.AppLog;
-import io.avaje.http.client.HttpClient.GeneratedComponent;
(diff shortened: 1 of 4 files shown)


http-client/src/main/java/io/avaje/http/client/DHttpApi.java
```diff
@@ -34,16 +32,22 @@ <T> void addProvider(Class<T> type, HttpApiProvider<?> apiProvider) {
   }
 
   @SuppressWarnings("unchecked")
-  <T> T provideFor(Class<T> type, HttpClient httpClient) {
-    final var apiProvider = (HttpApiProvider<T>) providerMap.get(type);
+  <T> T provideFor(Class<T> type, HttpClient httpClient, ClassLoader classLoader) {
+    var apiProvider = (HttpApiProvider<T>) providerMap.get(type);
+
+    if (apiProvider == null) {
+      init(classLoader);
```
rbygrave, line 39: I didn't really like this mutation here with init() ...

... produced 607 as an alternative approach.

Issue #539, opened 2025-01-11
Add support for path custom type conversion


SentryMan:
From the discord:

![image](https://github.com/user-attachments/assets/93fba763-abcc-4829-8d18-3394271faea8)

I'm thinking we could make it so that: 

```java
@Get("/something")
public void getMetadata(@QueryParam("id") Foo foo) {
  //
}
```

could generate something like:

```java
    var id = asType(ctx.pathParam("id"), Foo::new);
    var result = controller.getMetadata(id);
```

Where `asType` would be a new method added that takes a reference to a constructor/valueOf method

rbygrave:
Can we get more concrete examples of the types for this use case rather than Foo? I presume these are DDD types but let's get some good understanding of the motivation.

There is some interaction with the logic that determines what parameter is used for the body content. This is probably the main concern here - that we don't suffer in terms of handling the body parameter. This is somewhat because as avaje-http uses annotation processing we actually know the method parameter names and as such we then don't need explicit annotations for path parameters and query parameters ... and this then impacts the determination of which parameter is for the body.

Rightly or wrongly, we have the `@BodyString` annotation for the rare use case where the body is just a String. Calling that `@BodyString` instead of `@Body` almost feels like a mistake right at this moment.

SentryMan:
>  This is probably the main concern here - that we don't suffer in terms of handling the body parameter.

this is simple to handle, this should only apply if it's a GET/DELETE HTTP verb, or if the parameter is explicitly annotated with @QueryParam/@Header/Other

SentryMan:
> Calling that @BodyString instead of @Body almost feels like a mistake right at this moment.

I'm indifferent, but since we're jumping major versions to 3.0 you have the opportunity to change it.

rbygrave:
> Can we get more concrete examples of the types for this use case rather than Foo? I presume these are DDD types but let's get some good understanding of the motivation.

We still need this. The description uses Foo which does not actually express the **_motivation_**.

In addition the discord example _is not valid_ as it uses `{foo}` as a path parameter.

Ideally we get the description of this issue updated so that we clearly understand the motivation and use case.

SentryMan:
Closing until further details are provided

PR #644, opened 2025-09-25
Support custom type mapping

### LINKED ISSUE #640: Custom path and query parameter type conversion
Noting #539 was closed with a preference for some concrete examples, I have been migrating an existing REST backend to Avaje and noticed two situations where it would have been helpful to have the ability to provide a custom converter for path and query parameters:

1. The use of `Ulid` (from https://github.com/f4b6a3/ulid-creator/tree/master). The workaround is treat it as a string and convert in the method. Alternately it would be ideal if there was some way to detect common open source libraries people are using and provide the conversion in-place within the generated source.
2. The use of enums which are not encoded as strings compatible with `Enum.valueOf(Class<T extends Enum>, String)`. This is a more significant limitation as you need to either pass the enum values as strings for your own `static Enum from(String value)` method or refactor to a JSON type with `POST` so you can access something like Avaje JsonB (which is what I did). It is particularly unfortunate if the enum is a `@Json` anyway  (eg with `@Json.Value`) as it should prefer to use that as you already have an idiomatic string to enum mechanism already in place via the Avaje ecosystem.

As an aside it would have been good to work around the second scenario via a `@Filter` by way of replacing the query parameters, but it doesn't seem possible to replace the request passed through the chain (which is what the servlet API offers and would have for example enabled normalisation of the query parameters obtained from `Context.queryParams(String)` with something that would be `Enum.valueOf(..)` compatible).

SentryMan:
Adds a new annotation to register/import type handlers for custom types.

I'm not sure about the annotation name.

- adds new annotation for mapping
- if the type has no mapping and we know it is not a body (it's annotated by `@Header` or something) and it also have a constructor that takes a string, a mapping is generated
- you can import external types
- you can use static factory methods 

you can use it like:

```java

  @MappedParam
  @MappedParam.Import(Simple2.class)
  record Simple(String name) {}

  record Simple2(String name) {}

  @Form
  @Get("/typeForm")
  String typeForm(Simple s, Simple2 type) {
    return type.name();
  }

  @MappedParam(factoryMethod = "build")
  record Static(String name) {
    static Static build(String name) {
      return null;
    }
  }

  @Get("/typeFormParam")
  String typeFormParam(@FormParam String s, @FormParam Static type) {
    return type.name();
  }

  @Get("/typeQuery2")
  String typeMultiQuery(@QueryParam @Default({"FFA", "PROXY"}) Set<Simple> type) {
    return type.toString();
  }

  record Implied(String name) {}

  @Post("/typeQueryImplied")
  String typeQueryImplied(String s, @QueryParam Implied type) {
    return type.name();
  }
```

resolves #640  and https://github.com/avaje/avaje-http/issues/539

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/MappedParam.java
+++ b/http-api/src/main/java/io/avaje/http/api/MappedParam.java
@@ -1,0 +1,30 @@
+package io.avaje.http.api;
+
+import static java.lang.annotation.ElementType.MODULE;
+import static java.lang.annotation.ElementType.PACKAGE;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Marks a type to be mapped. */
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.TYPE})
+public @interface MappedParam {
+
+  /** Factory method name used to construct the type. Empty means use a constructor */
+  String factoryMethod() default "";
+
+  @Retention(SOURCE)
+  @Target({TYPE, PACKAGE, MODULE})
+  @interface Import {
+
+    Class<?> value();
+
+    /** Factory method name used to construct the type. Empty means use a constructor */
+    String factoryMethod() default "";
+  }
+}
--- a/http-api/src/main/java/io/avaje/http/api/PathVariable.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathVariable.java
@@ -1,0 +1,21 @@
+package io.avaje.http.api;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/** Marks a method parameter to be a path variable. */
+@Retention(RUNTIME)
+@Target({PARAMETER, FIELD})
+public @interface PathVariable {
+
+  /**
+   * The name of the path variable.
+   *
+   * <p>If left blank the method parameter name is used.
+   */
+  String value();
+}
--- a/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
+++ b/http-api/src/main/java/io/avaje/http/api/PathTypeConversion.java
@@ -98,6 +98,12 @@ public final class PathTypeConversion {
     }
   }
 
+  /** Convert to type. */
+  public static <T> T asType(Function<String, T> typeConversion, String value) {
+    checkNull(value);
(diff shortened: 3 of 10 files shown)


http-api/src/main/java/io/avaje/http/api/MappedParam.java
```diff
@@ -0,0 +1,30 @@
...
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.TYPE})
+public @interface MappedParam {
+
+  /** Factory method name used to construct the type. Empty means use a constructor */
+  String factoryMethod() default "";
+
+  @Retention(SOURCE)
+  @Target({TYPE, PACKAGE, MODULE})
+  @interface Import {
```
rbygrave, line 23: I think we additionally could add a repeatable for `@MappedParam.Import` ... as we'd likely specify all those together in one place.

http-api/src/main/java/io/avaje/http/api/PathVariable.java
```diff
@@ -0,0 +1,21 @@
+package io.avaje.http.api;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/** Marks a method parameter to be a path variable. */
```
rbygrave, line 10: Javadoc needs to include some commentary on when this is needed etx.

http-api/src/main/java/io/avaje/http/api/PathVariable.java
```diff
@@ -0,0 +1,21 @@
...
+@Retention(RUNTIME)
+@Target({PARAMETER, FIELD})
```
rbygrave, line 12: FIELD??

http-api/src/main/java/io/avaje/http/api/PathVariable.java
```diff
@@ -0,0 +1,21 @@
...
+public @interface PathVariable {
+
+  /**
+   * The name of the path variable.
+   *
+   * <p>If left blank the method parameter name is used.
+   */
+  String value();
```
rbygrave, line 20: Not sure this should exist at all?

http-api/src/main/java/io/avaje/http/api/PathVariable.java
rbygrave: Hmm, not convinced this should exist at all. There is some design decision made here that isn't clear - no test uses PathVariable etc.

rbygrave (changes requested):

http-api/src/main/java/io/avaje/http/api/PathVariable.java
SentryMan: Beanparams

http-api/src/main/java/io/avaje/http/api/PathVariable.java
rbygrave: Ok, I think we remove this for now as it isn't used in any tests and doesn't look strictly required.  I'll push that now.

Issue #647, opened 2025-10-06
JEP 467: Enhanced Markdown Comments


mechite:
Discussion about support for this ongoing in the Discord thread:
[JEP 467: Enhanced Markdown Comments](https://discord.com/channels/1074074312421683250/1424232065284440188)

`avaje-http` reads Javadoc for OpenAPI generation.
- If we modify the test suite to have a markdown version, Avaje build toolchain requires Java 25+ then
- If we bump that requirement, doesn't it also make sense for Avaje to start using that kind of doc internally too?
    - I do not mind personally doing the boring refactoring to make all our Javadoc use the new format
    - It is possible to use Markdown comments on an older Java release.
        - You can even target Java 11 and use JDK 25 as your build toolchain, and use markdown comments.
        - The only things this affects is - 1)  cannot build Javadoc site without JDK 25 - 2) some people may have outdated IDEs?

Modifying the generator to support both Javadoc syntax should not be too difficult.
- We are only reading main Javadoc description - but I don't think the current solution supports e.g. `{@link }`?
- We only read that up-to the last period (`.`)?
- Stuff like `@return`, `@param` is unchanged. We just need to be able to support both `///` & `/**`.

<sup>that something is not complex doesn't mean it can't be a lot of work :-)</sup>

mechite:
> does it not work now?
Nope, so I think that should be an issue for that library then
Once it works, the test in that PR should pass, and I guess there is no other changes to make?

So it could just be merged afterwards

SentryMan:
no like, I'm testing it and it work

SentryMan:
> Nope, so I think that should be an issue for that library then

we pull it from the JDK, not a library

SentryMan:
Closing as we already support

PR #648, opened 2025-10-06
add JEP 367 support

mechite:
### Do not merge
(see #647)


tests/test-javalin-jsonb/src/test/java/io/avaje/http/generator/JavalinProcessorTest.java
```diff
@@ -177,6 +177,41 @@ public void testOpenAPIGeneration() throws Exception {
...
+        openAPIController = List.of(file);
+    }
+    final var compiler = ToolProvider.getSystemJavaCompiler();
+
+    final var task =
+        compiler.getTask(
+            new PrintWriter(System.out),
+            null,
+            null,
+            List.of("--release=11", "-AdisableDirectWrites=true"),
```
SentryMan, line 198: release 23 minimum if you want markdown

mechite: you can go ahead and change that - if you can get this test passing, I think it's good to have?
Up to you whether you think it's good to have a direct test for markdown

also, there is a direct copy of this in tests for sigma generation etc too

SentryMan: I mean the method we use to get the javadoc is the same for both non markdown and markdown so I don't really see the point

PR #670, opened 2025-11-20
Adds a default validation exception handler

### LINKED ISSUE #669: [enhancement] Default middleware for handling validation exceptions
`avaje-http` provides a validation API, where `avaje-validator` among others are available.
Having `avaje-http` automatically register, for example, the following, could add convenience:

```java
jex.error(ValidationException.class, (context, exception) -> {
      List<ValidationException.Violation> violations = exception.getErrors();

      int violationCount = violations.size();
      String violationList = violations.stream()
          .map(violation -> "'" + violation.getField() + "' " + violation.getMessage())
          .collect(joining("\n"));

      // return a plain-text error message
      context.write(format("Bad Request [{0} validation violations: \n{1}]", violationCount, violationList));
});
``` 

1. `avaje-http` provides an agnostic `io.avaje.http.api.ValidationException` among each validator implementation.

    Therefore, this can be used by the generator, and assumed to exist in all cases.
    If a user does not have a validator, the error-handling middleware is just a hidden no-op they won't notice.

2. The above example displays a plain-text error message.

3. When `avaje-jsonb` is available (or just `avaje-json-core` really):
[RFC 7807](https://datatracker.ietf.org/doc/html/rfc7807) / [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html) compliant message as follows can be emitted from the `ValidationException`:
    ```json
    {
     "type": "https://avaje.io/validator/validation-error-message",
     "title": "Your request is not valid.",
     "errors": [
                 {
                   "detail": "must be a positive integer",
                   "pointer": "#/age"
                 },
                 {
                   "detail": "must be 'green', 'red' or 'blue'",
                   "pointer": "#/profile/color"
                 }
              ]
    }
    ```
    (this is a really bad, rough example)

4. Javalin offers a default middleware for their `Validator` as well.
    ```java
    app.get("/test123/{value}", ctx -> {
        int value = ctx.queryParamAsClass("value", Integer.class)
            .check(q -> (q > 0), "Quantity must be positive")
            .get();
        ctx.result(value);
    });
    ```

    Below is what their middleware looks like:
    ```java
    app.exception(ValidationException::class, (error, ctx) ->
        ctx.json(e.errors).status(400);
    });
    ```

SentryMan:
Resolves #669 by adding a default handler via the inject plugin

### DIFF
--- a/http-api/src/main/java/io/avaje/http/api/StreamingOutput.java
+++ b/http-api/src/main/java/io/avaje/http/api/StreamingOutput.java
@@ -1,0 +1,24 @@
+package io.avaje.http.api;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * An avaje {@link Controller} endpoint is able to use an instance of this interface
+ * as a return type.
+ *
+ * <pre>{@code
+ * @Post("some_endpoint")
+ * @Produces("application/octet-stream")
+ * public StreamingOutput thumbnail(
+ *   InputStream inputStream,
+ *   @QueryParam(Constants.KEY_SIZE) @Min(1) @Max(Constants.MAX_SIZE) Integer size
+ * ) throws IOException {
+ *   return (os) -> os.write(new byte[] { 0x01, 0x02, 0x03 });
+ * }
+ * }</pre>
+ */
+
+public interface StreamingOutput {
+  void write(OutputStream outputStream) throws IOException;
+}
--- a/http-client/src/main/java/io/avaje/http/client/DHttpApi.java
+++ b/http-client/src/main/java/io/avaje/http/client/DHttpApi.java
@@ -42,7 +42,8 @@ final class DHttpApi {
               + type
               + "\nPossible Causes: \n"
               + "1. Missing @Client or @Client.Import annotation.\n"
-              + "2. The avaje-http-client-generator dependency was not available during compilation\n");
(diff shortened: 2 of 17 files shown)


mechite:
d55a7a6 - OK, but we better verify that this is indeed valid for the `tag:` URI

rbygrave:
I'm not convinced by the "packaging" of where this should be yet - for example the Helidon implementation probably should go into avaje-nima as that is the opinionated combination of validator + inject + helidon + jsonb.

This has to rely on optional dependencies and reflection which makes me go "Hmmm".

SentryMan:
it's no worse than the [automatic validator registration ](https://github.com/avaje/avaje-validator/blob/2a04f2a3f000fd18c9b187f1f9ba45491709d732/validator-http-plugin/src/main/java/io/avaje/validation/http/HttpValidatorProvider.java#L18) for http

SentryMan:
you know I forgot you could catch those, time to replace every `Class.forName`

SentryMan:
> Helidon implementation probably should go into avaje-nima as that is the opinionated combination 

What of the people that want to use hibernate or a custom implementation? This PR is supposed to be agnostic and not favor any specific validator implementation.

> This has to rely on optional dependencies and reflection which makes me go "Hmmm".

Reflection is gone now, and we even have an optimized path through JPMS lookup when running on the module path. What downsides do you see of optional dependencies?

rbygrave (approved):

PR #674, opened 2025-11-22
Support for Jax-RS style "StreamingOutput" for JEX

andponlin:
This PR only implements for the JEX server option; I have not done this for the other server options. I tried to write a unit test for this, but it is quite difficulty with the visibility of types and methods.


andponlin:
> you could try to write a test in the test-jex section. did that not work?

I started to write a unit test, but the mocking was very difficult and it was quite dependent on statically initialized types. I just noticed the `test` module so will TAL a the options to implement a test there.

PR #686, opened 2025-12-03
Use Avaje Prism Utils

SentryMan:


http-generator-client/src/main/java/io/avaje/http/generator/client/ComponentMetaData.java
```diff
@@ -28,14 +28,15 @@ String fullName() {
...
 
       var defaultPackage =
-        !topPackage.contains(".")
-          && APContext.getProjectModuleElement().isUnnamed()
-          && APContext.elements().getPackageElement(topPackage) == null;
+          !topPackage.contains(".")
+              && APContext.getProjectModuleElement().isUnnamed()
+              && APContext.elements().getPackageElement(topPackage) == null;
 
+      System.err.println("ComponentMetaData detected top package: " + topPackage + " defaultPackage:" + defaultPackage);
```
rbygrave: We want this System.err.println() ?

PR #708, opened 2026-01-12
Merge multiple OpenAPI definitions

### LINKED ISSUE #51: Adding endpoint from other libs to openapi documentation

cbarlin:
Resolves #51

I've marked this as draft in case you want to change the approach taken, and because I've yet to fully finish the testing. The remaining changes are:

 * Add another `tests` module that builds off of `test-javalin-jsonb` so we can check it correctly pulls in from libraries
 * Adjust the github pipelines to ensure that the plugin is being used

I'm also not sure if this should be a separate goal or just part of the existing one.

I was also thinking possibly (as a second PR) of moving the json serialisation of the OpenAPI spec to a library, and the `http-generator-core` can use that to serialise/deserialise

Another thing I thought of as I was working on this is that this change works for libraries that have OpenAPI definitions, but not for ones in which there's no javadoc available for the annotation processor to parse... Is it worth adding the ability to parse `Schema` annotations as well?


rbygrave:
This PR is still DRAFT per se ... but yeah I'm pretty happy with it.

> What are your thoughts on making the plugin versioned alongside the rest of the project?

Maybe sort out this PR first?

> We can then also spin the jsonb stuff for the OpenAPI schema into a seperate module and then the http-generator-core can use it too.

Yes, could be a good followup.

tests/test-jex/src/main/resources/public/openapi.json
cbarlin: So turns out this isn't up-to-date on `master` either. I could make the github tests confirm that the git repo is clean (so the tests fail if these generated files aren't up-to-date)?

This file is semantically identical to the one in `master` if you regenerate it.

rbygrave (approved):

cbarlin:
Awesome - thank you! 

I'll get started on those follow-up items soon-ish 😁  Might also take a look at fixing #709 too - I'm pretty sure it's due to the part that handles generics for lists and sets etc. not including `Optional`.

PR #746, opened 2026-03-18
Add support for parsing `@Schema` annotations

cbarlin:
Adds support for reading (some) values out of an `@Schema` annotation and using that to populate the OpenAPI that's generated.

My biggest use-case is that I have records that are in another compilation unit, so reading the Javadoc doesn't work to get the descriptions. I also have a few `@Json` annotated items that have `@Json.Value` (retention of `SOURCE`, not available to other compilation units) and so needed a way to override the implementation. The other items (`example`, `default`, `title`, `name`, `pattern`) that are read seemed simple/harmless enough to parse out and add if someone desired.

There is also the scenario in which someone has both Javadoc and an `@Schema` annotation with a description in it... I decided that I'd make the annotation win, just in case the Javadoc was to indicate something else (though if the description isn't in the `@Schema` tag it should keep the Javadoc one)

### DIFF
--- a/aws-cognito/http-client-authtoken/pom.xml
+++ b/aws-cognito/http-client-authtoken/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>io.avaje</groupId>
       <artifactId>avaje-http-client</artifactId>
-      <version>3.8-RC4</version>
+      <version>3.8-RC5</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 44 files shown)


http-generator-core/src/main/java/io/avaje/http/generator/core/openapi/SchemaDocBuilder.java
```diff
@@ -277,14 +293,20 @@ private String getObjectSchemaName(TypeMirror type) {
   private <T> void populateObjectSchema(TypeMirror objectType, Schema<T> objectSchema) {
     Element element = types.asElement(objectType);
     for (VariableElement field : allFields(element)) {
-      Schema<?> propSchema = toSchema(field.asType());
+      Schema<?> propSchema = SchemaPrism.getOptionalOn(field)
+        .map(SchemaPrism::implementation)
+        .filter(typeMirror -> !"java.lang.Void".equals(typeMirror.toString()))
```
cbarlin: I should extract reading the `implementation` to a helper function.

PR #762, opened 2026-05-02
Add `@Body` to resolve ambiguous scalar/collection parameter inferenc…

rbygrave:
…e. For example, `List<String>` being treated as query params instead of request body.

This new `@Body` effectively replaces and deprecates the existing `@BodyString`, as it handles String type but also the other scalar types and collections of scalar types. Code should migrate to use `@Body`.

This also includes a fix to the generator behavior for explicit BODY collections. ElementReader now avoids applying query/form collection conversion wrappers to BODY params. That is, incorrect generated code when @Body is used on collection/scalar-container types.

Example:

```java
  @Post("postListParams")
  String postListParams(@Body List<String> codes) {
    return "codes:" + codes;
  }
```
... when we want `List<String> codes` via request body and NOT [inferred as] query parameters.

### DIFF
--- a/http-generator-jex/src/main/java/io/avaje/http/generator/jex/JexAdapter.java
+++ b/http-generator-jex/src/main/java/io/avaje/http/generator/jex/JexAdapter.java
@@ -49,31 +49,15 @@ class JexAdapter implements PlatformAdapter {
 
   public static String writeJsonbType(UType type) {
     var writer = new StringBuilder();
-    switch (type.mainType()) {
-      case "java.util.List":
-        writeType(type.paramRaw(), writer);
-        writer.append(".list()");
-        break;
-      case "java.util.Set":
-        writeType(type.paramRaw(), writer);
-        writer.append(".set()");
-        break;
-      case "java.util.Map":
-        writeType(type.paramRaw(), writer);
-        writer.append(".map()");
-        break;
-      case "java.util.stream.Stream":
-        writeType(type.paramRaw(), writer);
-        writer.append(".stream()");
-        break;
-      default: {
-        if (type.mainType().contains("java.util")) {
-          throw new UnsupportedOperationException(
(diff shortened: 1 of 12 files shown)


http-generator-jex/src/main/java/io/avaje/http/generator/jex/JexAdapter.java
```diff
@@ -49,31 +49,15 @@ public String bodyAsClass(UType type) {
...
-        }
-        writeType(type, writer);
-      }
+    if (type.mainType().contains("java.util")
+      && !type.mainType().startsWith("java.util.List")
+      && !type.mainType().startsWith("java.util.Set")
+      && !type.mainType().startsWith("java.util.Map")
+      && !type.mainType().startsWith("java.util.stream.Stream")) {
+      throw new UnsupportedOperationException(
+        "Only java.util Map, Set, List and Stream are supported JsonB Controller Collection Types");
```
SentryMan, line 58: why was this change made?

rbygrave: I'd say that JexAdapter.writeJsonbType() has always been a bit broken.

It's only called for `bodyAsType()`, so it needs to return a `Type` and so the `.list()`, `.set()`, `.map()`, `.stream()` etc are actually invalid for this case (they are for a JsonType<T> and not a JDK Type).

For `List<String>`, type.paramRaw() is just String, and writeType() expects a parameterized type, so it emits the incorrect code of:

`var codes = ctx.<List<String>>bodyAsType(Types.newParameterizedType(String.class, .class).list());`

When really it should be:

`var codes = ctx.<List<String>>bodyAsType(Types.newParameterizedType(List.class, String.class));`

SentryMan (approved):

PR #763, opened 2026-05-05
Generate $RequestFactory in avaje-http for request-scoped controllers

rbygrave:
### Summary

 Move request-scoped controller factory generation from avaje-inject into avaje-http. For controllers that inject request-scoped types (e.g. Context, ServerRequest, ServerResponse), avaje-http now generates a $RequestFactory class alongside the existing $Route. The $Route uses $RequestFactory instead of the avaje-inject-generated $Factory.

### Motivation

 See avaje/avaje-inject#1017. avaje-inject currently contains HTTP-specific machinery (BeanFactory, BeanFactory2, RequestScope, BeanRequestParams) that exists solely to support request-scoped controllers. This couples avaje-inject to HTTP server concepts (Javalin Context, Helidon ServerRequest/ServerResponse, etc.) which arguably belongs in avaje-http instead.

### Design choice

 Rather than removing the existing $Factory generation from avaje-inject (which would break older avaje-http versions), this uses a different name — $RequestFactory — so both can coexist:

 - avaje-inject continues generating $Factory as before (backward compat, unused by new avaje-http)
 - avaje-http now generates $RequestFactory — a plain @Singleton/@Component bean with @Inject fields for DI dependencies and a create(...) method for per-request controller construction
 - No coordination required between project versions — old or new avaje-inject works with new avaje-http

### Future phase

In a future phase, avaje-inject can deprecate and remove BeanFactory/BeanFactory2, RequestScope, BeanRequestParams, and the $Factory generation path entirely.

### What changed

 New: RequestFactoryWriter in http-generator-core — generates $RequestFactory for request-scoped controllers.

 Modified:
 - ControllerReader — now reads constructor params and classifies fields as request-scoped vs DI
 - PlatformAdapter — added default methods requestFactoryCreateParams() and requestFactoryImportTypes()
 - BaseProcessor — triggers RequestFactoryWriter when controller is request-scoped
 - All platform adapters (Jex, Javalin, Helidon, Vertx, Sigma) — implement the new platform methods
 - All platform ControllerWriter classes — reference $RequestFactory instead of $Factory

### Example

 For a request-scoped controller:

 ```java
@Controller 
class ReqScopedController { 

   @Inject Context context; // ... 
} 
```

 avaje-http now generates:

 ```java
@Singleton 
public final class ReqScopedController$RequestFactory { 

    public ReqScopedController create(Context ctx) { 
       var bean = new ReqScopedController(); 
       bean.context = ctx; 
       return bean; 
    } 
}
```

 And the $Route injects ReqScopedController$RequestFactory instead of ReqScopedController$Factory.
 
 ---------
 
 ### Why not `@AssistFactory/@Assisted`?

 We considered using assisted injection mechanism, where the controller would be annotated `@AssistFactory` with `@Assisted` on request-scoped params. This was ruled out because:

 1. Annotations can't be added to user classes by a processor — `@AssistFactory(SomeFactory.class)` and `@Assisted` need to be on the user's `@Controller` class. avaje-http's annotation processor generates new files but can't modify the user's source. The user would need to add these annotations manually, which is new boilerplate.
 2. The factory interface doesn't exist until code generation — @AssistFactory references a factory interface that the user defines. Here, the factory interface would need to be generated, creating a circular dependency: the user's class references a type that doesn't exist until compilation.
 3. Just generating the factory implementation seemed easier


SentryMan:
Add some tests and we're golden

rbygrave:
FYI:
---------

Here's the coverage matrix:

 
 │ Pattern                                                     │ Existing Coverage                                   │
 │ Field-only @Inject Context (Jex)                            │ ✅ test-jex ReqScopedController                     │
 │ Field-only @Inject Context (Javalin)                        │ ✅ test-javalin ReqScopedController                 │
 │ Field-only @Inject ServerRequest + ServerResponse (Helidon) │ ✅ test-nima-jsonb FieldInjectController            │
 │ Field-only @Inject ServerRequest only (Helidon)             │ ✅ test-nima-jsonb FieldInject2Controller           │
 │ Field-only @Inject ServerResponse only (Helidon)            │ ✅ test-nima-jsonb FieldInject3Controller           │

 │ Constructor mixed (DI + request-scoped)                     │ ❌ only in avaje-inject's inject-test (BController) │
 │ Constructor mixed with generics                             │ ❌ only in avaje-inject's inject-test (CController) │
 │ Field DI + method injection of request-scoped               │ ❌ only in avaje-inject's inject-test (AController) │
 │ Field DI + field request-scoped                             │ ❌ not tested in any avaje-http test                │


 ### Suggested tests to add

 1. Constructor mixed (DI dep + request-scoped param) — e.g. in test-jex or test-nima-jsonb:
   ```java
     @Controller
     class ConstructorReqScopedController {
         final MyService service;
         final Context context;
         ConstructorReqScopedController(MyService service, Context context) { ... }
     }
   ```
   This exercises the diConstructorParams path in RequestFactoryWriter.
 2. Field DI + field request-scoped (mixed fields) — e.g. in test-jex:
   ```java
     @Controller
     class MixedFieldController {
         @Inject MyService service;
         @Inject Context context;
     }
   ```
   This exercises both diFields and requestScopeFields in RequestFactoryWriter.
 3. Constructor mixed with generic DI dependency (Helidon) — e.g. in test-nima-jsonb:
   ```java
     @Controller
     class GenericDiController {
         final SomeService<Foo, Bar> service;
         final ServerRequest request;
         final ServerResponse response;
         GenericDiController(SomeService<Foo, Bar> service, ServerRequest request, ServerResponse response) { ... }
     }
   ```
   This exercises the generic type handling in imports and field declarations.
 4. Method injection of request-scoped type — e.g. in test-jex:
   ```java
     @Controller
     class MethodInjectController {
         @Inject MyService service;
         Context context;
         @Inject void setContext(Context context) { this.context = context; }
     }
   ```
   This is the known gap — RequestFactoryWriter doesn't handle inject methods yet. Adding this test would document/expose the gap.

 ### Recommendation

 Tests 1 and 2 are most important — they cover constructor-based DI deps which is the common real-world pattern and is already supported by the implementation. Test 3 validates generic edge
 cases. Test 4 would fail today and could be deferred or tracked as a known limitation.

SentryMan:
you don't have to post everything the AI responds with

SentryMan (approved):

rbygrave:
> you don't have to post everything the AI responds with

Yeah, I thought it was both interesting and useful in this case.

PR #766, opened 2026-05-14
Add http client otel module, Open Telemetry support for HttpClient

rob-bygrave:


SentryMan (approved):
Are people asking for this?

rbygrave:
Me, I'm wanting it. Partly because I'm seeing real success with graalvm native image (so the java agent approach doesn't work there). 

edit: we are in the process of migrating to grafana cloud [from DD].

PR #785, opened 2026-06-30
Add TestAPI client generation for "Contract-First" controllers

rob-bygrave:
When a server controller implements a contract-first interface, it currently does not generate a *TestAPI client, and this thus it's harder to write tests with access to the response status code and headers etc.

This modifies the TestClientWriter, to also generate the TestAPI for those controllers that implement a contract-first interface.


rbygrave:
@SentryMan  I'm keen on getting this one in, and doing a RC relesse.

SentryMan (approved):

PR #793, opened 2026-07-30
Fix for @Default ignored for "contract-first" style controller

rbygrave:
Given an interface with a `@Default` on a `@QueryParam` like:
```java
@QueryParam("useMaster") @Default("false") boolean useMaster
```
The `@Default("false")` was not being read and applied on the query parameter. This resulted in a null being used and a InvalidPathArgumentException.

```
io.avaje.http.api.InvalidPathArgumentException: path element is null
 at io.avaje.http.api.PathTypeConversion.checkNull(PathTypeConversion.java:52)
 at io.avaje.http.api.PathTypeConversion.asBoolean(PathTypeConversion.java:178)
 at io.avaje.http.api.PathTypeConversion.asBool(PathTypeConversion.java:186)
 ...
```

#### The fix has 2 parts:
A) Improve ProcessingContext.superMethods() which only matched on name and instead match also on parameter types etc 
B) Fix MethodReader to use the annotated element [from the super method / interface] as the controller method itself does not have those annotations like `@Default`.

--------------
In the generated code, with the bug we see the null value like:
```java
var useMaster = asBool(queryParams.contains("useMaster") ? queryParams.get("useMaster") : null);
```
... when we expect the default value of "false" like:
```java
var useMaster = asBool(queryParams.contains("useMaster") ? queryParams.get("useMaster") : "false");
```

---------------

## Workaround

Copy the annotations from the interface (like `@QueryParam("useMaster") @Default("false")`) to the controller method. This duplicates the annotations so not ideal but does mean that the generated code includes the default values as expected.

### DIFF
--- a/http-generator-core/src/main/java/io/avaje/http/generator/core/MethodReader.java
+++ b/http-generator-core/src/main/java/io/avaje/http/generator/core/MethodReader.java
@@ -73,8 +73,7 @@ public class MethodReader {
 
     initWebMethodViaAnnotation();
 
-    this.superMethods =
-        superMethods(element.getEnclosingElement(), element.getSimpleName().toString());
+    this.superMethods = superMethods(element);
     superMethods.forEach(m -> methodRoles.addAll(Util.findRoles(m)));
     this.throwsList = element.getThrownTypes();
     this.hasThrows = !throwsList.isEmpty();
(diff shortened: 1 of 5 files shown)


SentryMan (approved):

rbygrave:
Retrospectively noting that this change introduced a regression for the "contract first" case that is resolved via: https://github.com/avaje/avaje-http/pull/802

PR #802, opened 2026-08-27
generator: Recover parameter names for contract-first controllers and…

rbygrave:
… run checkArgumentNames in MethodReader

When implementing an interface from a precompiled jar without -parameters, method parameter names from the interface default to arg0, arg1, etc. Fall back to the parameter names declared on the implementing controller method in source code, and run checkArgumentNames() during MethodReader.read() so all server generators (Helidon, Jex, Javalin) positionally match path segments from the route URL.

Add ContractControllerTest in test-nima-jsonb covering name-matched path params, swapped parameter orders, and query parameter binding.


rbygrave:
Note that this could be considered a "Regression Fix" to https://github.com/avaje/avaje-http/pull/793 that was introduced in version 3.11

That is, that change 
-  (A) read annotations from the interface, ... what we wanted for that change
-  (B) but **also read the parameter name from the interface**  ... which was a side effect that wasn't spotted and effectively a regression/change in behaviour

SentryMan (approved):

PR #812, opened 2026-09-11
Fix @MappedParam asType conversion code generation

ah3nan:
### Problem
Two bugs in the `@MappedParam` custom parameter-conversion code path:

1. **Missing separator in `asMethod()`** — `CustomHandler.asMethod()` did not append the `", "` that `toMethod()` has, so a *required* path param generated invalid code:
   ```java
   var TsID = asType(TsID::buildpathParams.contains("TsID") ? pathParams.get("TsID") : null);
   ```
   → `method asType ... cannot be applied to given types`.

2. **Type-name shadowing** — when the local variable name equals the type name (e.g. `TsID TsID`), the generated method reference resolves to the variable being declared:
   ```java
   var TsID = asType(TsID::build, ...);   // javac: cannot use 'var' on self-referencing variable
   ```

### Change
- `http-generator-core/TypeMap.java` — `CustomHandler.asMethod()` now appends the missing `", "` separator (matching `toMethod()`).
- `http-generator-core/ElementReader.java` — added `avoidTypeShadowing(...)`; when the parameter variable name equals the type's short name, the method reference is qualified with the fully qualified type.

### Verification
With `@MappedParam(factoryMethod = "build") record TsID(String value)` and `@Get("x/{TsID}") String m(TsID TsID)`, the generated handler is now:

```java
var TsID = asType(org.example.TsID::build, pathParams.contains("TsID") ? pathParams.get("TsID") : null);
```

and compiles cleanly. `test-nima-jsonb` (which exercises `@MappedParam`) passes.


SentryMan (approved):

rbygrave:
WTF another PR merged without approval @SentryMan ?

PR #816, opened 2026-09-19
Fix Helidon parameter binding for empty values

rob-bygrave:
Use Parameters.first() instead of contains() plus get() when generating scalar
 path, query, form, and cookie bindings. This prevents empty submitted values from
 causing IndexOutOfBoundsException in Helidon ParametersMap.

 Before:
     `formParams.contains("name") ? formParams.get("name") : null`

 After:
     `formParams.first("name").orElse(null)`

 An empty form submission such as `name=` could produce:

     io.helidon.http.RequestException: Internal Server Error
     Caused by: java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
         at io.helidon.common.parameters.ParametersMap.get(ParametersMap.java:77)
         at ...IndexController$Route._searchOrganisations(IndexController.java:55)

 Note that originally the generated code did use `.first("name").orElse(null)` but was changed to the contains + get approach for performance (noted in the helidon docs around the extra cost of the optionals).

 So yes this change is the safe vs performance tradeoff that the helidon api has.


rbygrave:
Alright, need this change so merging in.

SentryMan (approved):

